PR #10, opened 2017-02-28
Updated the app to the latest sdk version 25 and added fabric.io Crashlytics

mehtamanan0:


build.gradle
```diff
@@ -1,25 +1,22 @@
...
-        }
+        jcenter()
+        maven { url 'https://maven.fabric.io/public' }
     }
 
     dependencies {
-        classpath "com.android.tools.build:gradle:1.3.0"
-        classpath "com.crashlytics.tools.gradle:crashlytics-gradle:1.14.7"
+        classpath "com.android.tools.build:gradle:2.2.3"
         classpath "com.jakewharton.sdkmanager:gradle-plugin:0.12.0"
```
zaclimon: You might want to remove the sdkmanager as the gradle tools take all the necessary prerequisites.

Source: 
https://github.com/JakeWharton/sdk-manager-plugin

mehtamanan0: Cool ! thanks will do that.

megabitdragon:
Please don't change the fabric API key. If you want to use your own fabric account, use it in your fork but don't send a pull request. Thank you.

mehtamanan0:
Umm i accidentally changed it but reverted it back in the second commit.

zaclimon:
@megabitdragon wouldn't it be easier to just reserve a space for the fabric API key? This way, the "official" release would have it's own and any user who would want to fork would insert their own key. (Simillar to how it is made in api.properties)

megabitdragon:
Good point @zaclimon. We should create a fabric.properties with the API key.

build.gradle
```diff
@@ -93,11 +90,11 @@ dependencies {
         maven {
             url "http://download.crashlytics.com/maven"
```
zaclimon: The Crashalytics link is no longer necessary as all it's stuff is in the Maven public repo.

mehtamanan0: ohh okay will remove that in the next commit.

PR #11, opened 2017-03-01
adds thumbnail images for image files

csoni111:
**This pr aims to implement feature [#2191](https://bugs.amahi.org/issues/2191). Check the screenshot below:**

thumbnailimages

### DIFF
--- a/src/main/java/org/amahi/anywhere/adapter/ServerFilesAdapter.java
+++ b/src/main/java/org/amahi/anywhere/adapter/ServerFilesAdapter.java
@@ -20,6 +20,7 @@
 package org.amahi.anywhere.adapter;
 
 import android.content.Context;
+import android.net.Uri;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
@@ -27,8 +28,12 @@ import android.widget.BaseAdapter;
 import android.widget.ImageView;
 import android.widget.TextView;
 
+import com.squareup.picasso.Picasso;
+
 import org.amahi.anywhere.R;
+import org.amahi.anywhere.server.client.ServerClient;
 import org.amahi.anywhere.server.model.ServerFile;
+import org.amahi.anywhere.server.model.ServerShare;
 import org.amahi.anywhere.util.Mimes;
 
 import java.util.Collections;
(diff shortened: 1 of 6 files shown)


cpg:
It works well! Great job!

My only complaint is that it's a little inefficient in that it loads the pictures once. As the preview advances, the whole picture is loaded and displayed in the thumbnail. Then when the user clicks on a picture, it loads the picture again. This is not a big problem locally in a LAN, but it matters over cell data.

There are two things that could help with this:

1. If the picture could be transparently cached as it's loaded first, and the app could get it straight from the thumbnail cache, it may just solve the issue
1. If the app and the server were to be using best practices for web protocols, we could be using If-Not-Match (Etag), If-Modified-Since, etc. so that a new request would return a [304 Not Modified](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/304) code and the re-download of the whole image (megabytes, these days) would be avoided.

Looking a bit more on our server (a beefed up version of this [server](https://github.com/amahi/spdy-proxy)), I believe it respects If-None-Match/ETag headers, but not If-Modified-Since. We probably should implement those, however!

Would you research this a bit more to see if there is some setting that could affect the app to take advantage of If-None-Match/Etag?

csoni111:
I digged deeper into the [docs ](http://square.github.io/picasso/2.x/picasso/) of picasso and found that whenever we create a new instance of picasso, it creates following:

```
The instance is automatically initialized with defaults that are suitable to most implementations.

LRU memory cache of 15% the available application RAM
Disk cache of 2% storage space up to 50MB but no less than 5MB. (Note: this is only available on API 14+ or if you are using a standalone library that provides a disk cache on all API levels like OkHttp)
Three download threads for disk and network access.
```

Also in this [stackoverflow answer](http://stackoverflow.com/questions/22804893/how-to-download-and-caching-bitmap-using-picasso-library#comment34800202_22805145) by Jake Wharton _(the developer of picasso)_ , he says that the `ETag` and `conditonal requests` are handled automatically by picasso.

So in short `picasso` along with `OkHttp` handles all the caching and other things automatically for us.

Moreover if required we can customize the size of the memory cache using Picasso.Builder(). Also we can provide a custom method to check Etag or If-Modified-Since headers by making a new HttpOkClient and providing it to picasso as in this [answer](http://stackoverflow.com/questions/40534963/make-picasso-to-play-well-with-headers-like-if-modified-since-and-etag).

So finally what should I do?

cpg:
I have to whip up a local server to test Etag. I am fairly sure we tested that feature in the server.

csoni111:
Similar to the issue #18 picasso loads outdated image from the cache without actually fetching from the server in one of the use case.
Consider an image with filename `example.jpg` on a server share, let it be fetched from the server once. Now on updating the image content on the server but keeping the filename same, picasso (checking that the url is same as previous) will loads the image from the memory cache instead of actually sending the request. This will result in new content not getting updated in the app.

csoni111:
As discussed with @cpg, a simple solution would be to add `mTime` (last modified time) param to the url of the image to be fetched from server which will change if the image content has changed. Hence changing the url.

cpg:
We did found one path where the If-None-Match/Etag was not generated properly in the back-end and [we fixed it, and released it](https://bugs.amahi.org/issues/2216).

cpg:
Please do add the change for cache-busting with mtime (all in small caps, if possible), and the removal of the okhttp caching.

csoni111:
@cpg done

PR #12, opened 2017-03-01
Move Crashlytics keys into a different file

zaclimon:
This seems enough but there is one issue I'm not quite sure how to tackle yet.

It seems that there is only one apiKey used at this time but there are two flavors. Is the same apiKey used for both flavors? (Beta and Release)

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,15 @@ Gradle, Android SDK and project dependencies will be downloaded automatically.
   client.id = ID
   client.secret = SECRET
   ```
+  
+  ```
+  $ vim fabric.properties
+  ```
+  ```
+  apiKey=YOUR_API_KEY
+  apiSecret=YOUR_API_SECRET (Optional)
+  ```
+
 
 3. Build the application.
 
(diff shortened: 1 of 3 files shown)


cpg:
Can you clarify what is the status of this PR?

We probably should keep the actual API key in private, like other files.

zaclimon:
The API key was stored in a separate metadata in AndroidManifest.xml. As suggested by Fabric, it is recommended that if Git is used to create a separate fabric.properties and then put the required keys there.

Then the user could keep it's key private by adding the file to his/her .gitignore.

Source:

https://docs.fabric.io/android/fabric/settings/working-in-teams.html#android-projects

Do you have some sort of special private file where you do keep all your API keys?

cpg:
I guess we could use the same api.properties, with different entries for the Crashlytics stuff.

Maybe that is best. Can you change this PR for that?

zaclimon:
Normally the library either checks the API key in the AndroidManifest.xml or fabric.properties. I guess it should be possible for Gradle to build dynamically a fabric.properties based on the contents of api.properties.

I'll take a look at this.

cpg:
Stale

Issue #19, opened 2017-03-05
Enhancement: Password field visibility


top-gun007:
**Summary**:
There should be a button to view typed in password just like in all popular apps.

**Screenshot:**
![image](https://cloud.githubusercontent.com/assets/15179401/23587193/05648784-01cd-11e7-8186-5a2611d0ab12.png)

asifali1234:
Hi all,
I would like to contribute to the Amahi Android app. I will try fixing this issue.

DoMINAToR98:
Can i fix this?

top-gun007:
I have assigned this to myself
Thank you

Get Outlook for Android<https://aka.ms/ghei36>



On Mon, Mar 6, 2017 at 3:24 AM +0530, "Kartik Sharma" <notifications@github.com<mailto:notifications@github.com>> wrote:


Can i fix this?

mehtamanan0:
This Feature is addded in the PR  #20 . @top-gun007 @asifali1234 @DoMINAToR98 Feel free to add any more functionality into the login screen as you may like , try to come up with Innovative ways to make the login Screen more Creative.

lakshyagupta21:
Is there any progress on this issue. If nobody is working on it then can i work on this ?

Ishmeetsingh97:
Hi, is there any developments on this ? I would like to work on it.

PR #20, opened 2017-03-05
adds show password functionality addressed in issue #19

mehtamanan0:
amahiswa

### DIFF
--- a/src/main/java/org/amahi/anywhere/activity/AuthenticationActivity.java
+++ b/src/main/java/org/amahi/anywhere/activity/AuthenticationActivity.java
@@ -24,6 +24,7 @@ import android.accounts.AccountAuthenticatorActivity;
 import android.accounts.AccountManager;
 import android.app.Activity;
 import android.os.Bundle;
+import android.support.design.widget.TextInputLayout;
 import android.text.Editable;
 import android.text.InputType;
 import android.text.TextWatcher;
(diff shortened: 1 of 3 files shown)


csoni111:
Just a suggestion: Following the material design guidelines, instead of a checkbox, `TextInputEditText` and `TextInputLayout` can be used given in the latest android support design library to solve issue #19. It supports password toggle feature by default.

Reference: [Material Guidelines](https://material.io/guidelines/components/text-fields.html#text-fields-password-input)

mehtamanan0:
hmm nice suggestion @csoni111  will do that!

cpg:
If I may, my main concern with this view, usability wise, is how the keyboard popping up covers all or most of the input fields. This happens with regular phones vertically.

 The logo should be either higher, smaller  (or both) or even under the whole set of input fields and the sign in button altogether. This way the keyboard would block the logo (and not the input fields) saving the user some awkward interactions to move field to field.

mehtamanan0:
For this issue we can address it either by using WindowSoftInputmode 
Refer here: https://android-developers.googleblog.com/2009/04/updating-applications-for-on-screen.html 

Or the other  option is to Change the complete layout itself.
@cpg what would you prefer?

top-gun007:
@mehtamanan0 I reported the issue first and it wasn't even approved so how come you went on to sending the PR I was planning on doing it.

top-gun007:
@cpg can you look into it please.

mehtamanan0:
@top-gun007 Sorry for that i will remove my PR.

PR #27, opened 2017-03-06
Added a new Preference Tell a friend.

octacode:
I have added a new Preference of tell a friend. This launches a shared Intent with the message "Amahi for your android. Download it from here " + playstore link.

screenshot_20170306-140351

OnPreferenceClick Listener :
screenshot_20170306-140358

### DIFF
--- a/src/main/java/org/amahi/anywhere/fragment/SettingsFragment.java
+++ b/src/main/java/org/amahi/anywhere/fragment/SettingsFragment.java
@@ -106,11 +106,13 @@ public class SettingsFragment extends PreferenceFragment implements Preference.O
 		Preference applicationVersion = getPreference(R.string.preference_key_about_version);
 		Preference applicationFeedback = getPreference(R.string.preference_key_about_feedback);
 		Preference applicationRating = getPreference(R.string.preference_key_about_rating);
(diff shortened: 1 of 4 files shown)


cpg:
cool!

what does this do. exactly what message does it send to the friend?

can you post the details of an email or whatever the other person (the new one) sees?

octacode:
@cpg on clicking it. A shared intent will open up giving the playstore link of the app. Here'a snippet of the new Gmail message that will be composed.
screenshot_20170306-182647

cpg:
The only problem is that to be able to use the app one needs to have an Amahi HDA, so installing the app without and HDA will only lead to frustration and possibly bad reviews.

I think the Subject field **has** to be populated, and the message has to be different. These are my suggestions

Subject: Check out my Amahi home server

Message: I use the Amahi home server for storing, backing up and streaming all my files. Check it out! https://www.amahi.org

octacode:
@cpg will update the message.

octacode:
**I have made the changes to the message.**
@cpg Please have a look.
screenshot_20170307-120352

octacode:
Thanks for the merge.

Issue #23, opened 2017-03-06
Use webview in settings's version


sukhbir-singh:
On selecting Version option from settings of the app, currently it is opening default browser and opening the web page. "Good practice" is to show the web page in the app only using web view, instead of redirecting it outside the app.

sukhbir-singh:
@cpg should i send the pull request corresponding to this issue.

PR #31, opened 2017-03-06
Resolved issue#23 Implemented webview in settings version option

sukhbir-singh:


cpg:
This fails to build with this error:

```
..../android-1/src/main/java/org/amahi/anywhere/activity/WebViewActivity.java
Error:(4, 30) error: package android.support.v7.app does not exist
Error:Execution failed for task ':compileDebugJavaWithJavac'.
> Compilation failed; see the compiler error output for details.
```

sukhbir-singh:
@cpg I have removed unnecessary import which is causing the problem. Now it should work fine, in case the problem persists please reply.

cpg:
fixes #23

Issue #46, opened 2017-03-08
deprecated android.support.v4.app.ActionBarDrawerToggle still in use


vjs3:
_android.support.v4.app.ActionBarDrawerToggle_ has been deprecated more than a year back and new android guidelines suggest us to use _android.support.v4.app.ActionBarDrawerToggle_. Above class and its few methods are still in use.

cpg:
merged already

PR #47, opened 2017-03-08
deprecated ActionBarDrawerToggle updated

vjs3:
As described in #46, use of _android.support.v4.app.ActionBarDrawerToggle_ is deprecated hence I have updated it with _android.support.v7.app.ActionBarDrawerToggle_ along with its methods as well. #46


megabitdragon:
Nice visual update.

Issue #21, opened 2017-03-05
Keyboard covering Up TextFields


mehtamanan0:
as @cpg mentioned there is a problem in some phones where the keyboard covers up the textfields , So what @cpg suggested was to move the logo down below like:

![screenshot 06-mar-2017 2_26_02 am](https://cloud.githubusercontent.com/assets/16348397/23591384/5803e484-01e7-11e7-8519-bbc320ceee2a.png)

What would you all suggest?

PiyushGupta007:
Can I work on it ?

mehtamanan0:
@PiyushGupta007  yes sure go ahead

lakshyagupta21:
Is it really necessary to change position of logo to just make textfield visible. Logo is always preferable at the top. The same thing can happen in some devices if you change orientation of your screen from portrait to landscape. To solve this issue what we can do move our layout  up whenver soft keyboard opens up. As our main priority is to write in textfield whenever softkeyboard opens up so other fields and images are not necessary to be visible at that time. Visible i mean if screen is small then we can't afford to make all component visible in that size of screen. Correct me if i am wrong anywhere.

csoni111:
I agree with @lakshyagupta21, moving the logo down is a bad idea!
As @mehtamanan0 suggested in [#20](https://github.com/amahi/android/pull/20#issuecomment-284251988), we can use `WindowSoftInputmode` in the manifest to move the layout up on opening the soft keyboard. That way the text fields will always be visible.

PiyushGupta007:
On it

cpg:
It's not just the text that needs to be visible, but the Sign In button as well, to avoid having to close the keyboard to be able to click it.

lakshyagupta21:
We can add event listener for enter key of soft keyboard which will move you to the "password" field if you are writing on "username" field and for "password" field enter key will directly call the login method. I don't think that we can afford to have two components (one EditText and Sign In Button) and also a soft keyboard at a time all together in devices with small screen size. Correct me if i am wrong.

octacode:
@lakshyagupta21 android handles this on its own. We just have to set softInputMode attribute in the manifest file as true.

pranavgupta1234:
Any updates on this ? I would like to resolve it :)

cpg:
I believe @PiyushGupta007 was working on this issue. Perhaps he can chime in with the status.

asifali1234:
[...]

PR #50, opened 2017-03-08
Resolved the issue #21 of keyboard covering the TextFields.

hshrimali:
Resolved the issue #21. Now, it shows both the TextFields and the SignIn button too.


hshrimali:
I have resolved the conflicts. @cpg Now you can check.

PR #63, opened 2017-03-09
Added screen transitions.

### LINKED ISSUE #62: Addition of screen transitions
Right now we don't have any screen transitions between intents. I think it will be good to add a nice slide_in and slide_out transition effect during intent transitions. I would like to work on this issue.

octacode:
I have added a screen transition of fade in and fade out when using intents.
Fixes #62


megabitdragon:
This is a matter of personal preference. I will not merge until we discuss it internally.

cpg:
This seems to make things just a little slower, though I'm not sure.

I could go either way on this one. It does not look like a lot of code, in any case, so it can be reversed quickly if we have to. Unless there are any objections, I will merge it in a few days.

Issue #64, opened 2017-03-09
Showing file size and last modified date along with file name in the list


sukhbir-singh:
Working on this right now!!

octacode:
Can close this one.

PR #66, opened 2017-03-09
Added file details to list item i.e. fixed issue #64

sukhbir-singh:
Fixed issue #64 successfully

Screenshots of changes made :--


octacode:
@sukhbir-singh Nice Job! But don't you think we should just mention the Local time instead of showing the whole thing. I mean to say is "Fri Sep 26 18:10:56" instead of "Fri Sep 26 18:10:56 GMT+5:30 2014".

sukhbir-singh:
@octacode Thanks for the suggestions. I will incorporate the changes.

sukhbir-singh:
@cpg @megabitdragon Please review this...

cpg:
This is a good idea, however, the implementation is not great, in my opinion. It makes things look very crowded and it detracts from the experience.

I like it that the timestamp is left aligned and the size is right aligned.

However, this is secondary information. Here are my recommendations that in my view would improve this:
- timestamp and size in smaller font, so it does not take a lot of room
- both should be in much less brighter color than the main details
- the timestamp is a little too unfriendly, perhaps instead of `Thu Sep 11 11:43:15 PDT 2014` it should say `Thu Sep 11 2014` and the size should be `7.28MB` without a space.
- actually, the timestamp should not be converted to local time, i think, since the file is not local, it's in a server. It should display the date in the file, namely, `Wed Sep 10 2014`.

About the last point, I checked and the mtime of this demo file is `"Wed, 10 Sep 2014 18:43:15 GMT"` so it's GMT-adjusted time. I am not sure that was a good idea from the server -- since the clients could be in different time zones. This is somewhat unlikely the vast majority of the time, so I guess the proper fix should be for the server to return the actual file mtime local to the server. By converting to local device time I think we're not doing a good service to the system, adding one more wrinkle to cover a bug.

----> So I would say use the mtime sent by the server, without showing GMT or anything. Later we can fix this server bug and send the local server time and that will just work, without changes to the client.

I hope that reasoning was clear.

sukhbir-singh:
@cpg  I will do the required modifications

sukhbir-singh:
This is the modified layout. Please review again.

csoni111:
I would like to give some suggestions here:

- Instead of keeping the date below the image, we can keep it like this

sample-image

- We may display Relative time like `Seconds ago`, `5 minute ago`, `3 Hours ago`, `4 days ago` and so on... but after a certain time limit we can change it to absolute date. Like for example after 10 days we can display exact date.

- @sukhbir-singh you have used `int` datatype to store the number of bytes of the file. Integer has a max limit of `2147483647` which is 1.99GB approx. So suppose if the file is greater than 2GB then the app may crash (I haven't tested it). You may change it to `long` datatype.

- Also you have only coded for file sizes in `MB`, what if the file size is in bytes (for example a simple `.txt` file of 200 bytes), it will look odd displaying `0.00MB`. Same goes for file size in `GB`.
You can create a separate function for this, which on passing size in bytes returns appropriate text in MB, KB, GB, or simply Bytes.

src/main/java/org/amahi/anywhere/server/model/ServerFile.java
```diff
@@ -43,7 +43,18 @@
 	@SerializedName("mtime")
 	private Date modificationTime;
 
-	public void setParentFile(ServerFile parentFile) {
+       @SerializedName("size")
+       private int size;
```
csoni111: Change it to `long` datatype.

src/main/res/layout/view_server_file_item.xml
```diff
@@ -46,4 +49,35 @@
...
+        android:paddingRight="8dp"
+        android:layout_width="match_parent"
+        android:layout_height="match_parent">
+
+        <TextView
+            android:textColor="#686868"
+            android:layout_weight="1"
+            android:id="@+id/last_modified"
+            android:maxLines="1"
+            android:textSize="@dimen/text_micro"
```
csoni111: This size is in `dp` please change it to `sp`.

src/main/java/org/amahi/anywhere/adapter/ServerFilesAdapter.java
```diff
@@ -104,8 +109,28 @@ private View newView(ViewGroup container) {
...
+               int size=getFileSize(file);
+
+               if(size==0){
+                 moreInfo.setVisibility(View.GONE);
+               }else{
+
+                 moreInfo.setVisibility(View.VISIBLE);
+
+                 double inMb=(size/1024.0)/1024.0;
+                 fileSize.setText(((int)(inMb*100))/100.0+"MB");
```
csoni111: This is a really bad way of conversion. You could use java's `String.format()` function to return only two decimal places.

src/main/java/org/amahi/anywhere/adapter/ServerFilesAdapter.java
```diff
@@ -104,8 +109,28 @@ private View newView(ViewGroup container) {
...
+		TextView fileSize = (TextView) view.findViewById(R.id.file_size);
+		TextView fileLastModified = (TextView) view.findViewById(R.id.last_modified);
+               LinearLayout moreInfo = (LinearLayout) view.findViewById(R.id.more_info);
 
 		fileTextView.setText(getFileName(file));
+
+               int size=getFileSize(file);
+
+               if(size==0){
+                 moreInfo.setVisibility(View.GONE);
```
csoni111: This results in a minor bug. I am not sure if I should call this a bug or not!. Well if the file size is 0 (that is in case of a newly created file). It hides the `date` also. 
I think you must not hide it, instead display both date and size. The size can be displayed as `0 Bytes`.

sukhbir-singh: @csoni111 I think for distinguishing a single file type from a folder, size can be used as a measure. Also displaying 0B below a folder doesn't seem reasonable. So may be hiding size is better than displaying size as 0B below folder name.

src/main/java/org/amahi/anywhere/adapter/ServerFilesAdapter.java
sukhbir-singh: Thanks for recommendation @csoni111, I will do this change.

src/main/java/org/amahi/anywhere/adapter/ServerFilesAdapter.java
cpg: folders (directories) have a specific mime type of `text/directory` the size should not be used as an indication of an entry being a folder.

sukhbir-singh:
@cpg  @csoni111 Review the changes....

cpg:
folders (directories) have a specific mime type of `text/directory` the size should **not** be used as an indication of an entry being a folder.

csoni111:
@sukhbir-singh check this screen:

device-2017-03-16-094615

It contains several images which are less than 10KB. So as you have only coded for MB, it displays 0.00MB, but instead it should display 5.14KB or something like that!

Also the `new_file.txt` is a newly created file with no text (hence 0 size), so it hides the info layout as per your if else check in the code. I think it must display the mtime as well as the size as 0B for newly created files too.

Also no need to display the size for a folder/directory but I think mtime can be displayed for folders too and as @cpg suggested you can check if a file is a directory or not using its mime type.

src/main/java/org/amahi/anywhere/server/model/ServerFile.java
```diff
@@ -43,7 +43,18 @@
 	@SerializedName("mtime")
 	private Date modificationTime;
 
-	public void setParentFile(ServerFile parentFile) {
+       @SerializedName("size")
+       private long size;
+
+       public void setSize(int size) {
```
csoni111: change this to `long` datatype too or you may remove this method as it is not required.

csoni111:
Also @sukhbir-singh pls add the `size` parameter to parcelable content that is in this constructor on [line#102](https://github.com/amahi/android/pull/66/files#diff-0fd37eb0ebeacce75187d653c15555a2R102) and in writeToParcel() method on [line#110](https://github.com/amahi/android/pull/66/files#diff-0fd37eb0ebeacce75187d653c15555a2R110).

Without these it will result in Null pointer error while reading the file size when fetching the file list from Bundle state at [line#287](https://github.com/amahi/android/blob/master/src/main/java/org/amahi/anywhere/fragment/ServerFilesFragment.java#L287).

cpg:
Are there no common libraries for easy human-readable size representation [like what we have in RoR](http://api.rubyonrails.org/classes/ActionView/Helpers/NumberHelper.html#method-i-number_to_human_size)?

sukhbir-singh:
Thank you for the hint @cpg  
I just find out [Formatter.formatFileSize](https://developer.android.com/reference/android/text/format/Formatter.html) useful in android, will send the modified code soon.

sukhbir-singh:
@csoni111 @cpg @megabitdragon I have made all the required changes needed. Please review and suggest if any other change needed.

cpg:
Great work!

PR #68, opened 2017-03-10
Adds swipeRefreshLayout in side nav bar, fixes #60

### LINKED ISSUE #60: No `RETRY` button on hda selection screen
hda_selection_screen

There is no retry button present in this side navigation bar. If the fetching of hda list from the server fails for some reason, there is only one option that is to close the app and reopen again.

This happens a lot with me and it is irritating. So I would like to implement this feature. Suggest if a `Button` should be added or `SwipeRefreshLayout`.

csoni111:
A swipe down to refresh layout has been implemented in the side navigation bar (the one which shows the spinner for hda selection). It will ease the process of refreshing the server list as discussed in #60.


cpg:
Fixes #60

PR #70, opened 2017-03-10
Made some changes in XML layouts

peytonswe:


hshrimali:
Sorry to say but your max line conversion and hard coded string conversions are already sent as PR before.  You should have looked first.

peytonswe:
Oh, sorry. I'll try to do something else.

hshrimali:
It's okay.

Issue #75, opened 2017-03-11
Showing DialogFragment for selecting sorting type in folders


sukhbir-singh:
I am working on this...

PR #77, opened 2017-03-12
Added DialogFragment for selecting sorting order for files

sukhbir-singh:
Solved issue #75


sukhbir-singh:
@cpg @megabitdragon Please review this...

cpg:
I do not like this, as it adds extra UI taps for the user and the sorting symbol is, I feel, well understood even by most mainstream users.

Thanks for the suggestion, however, I think we'll pass on this for now.

PR #80, opened 2017-03-13
Password Visibility issue solved.

Ishmeetsingh97:
#19 Solved.


megabitdragon:
I thinks this was mentioned before. The checkbox is not required. We should use [TextInputLayout](https://developer.android.com/reference/android/support/design/widget/TextInputLayout.html)

Ishmeetsingh97:
Ok i will send a PR for that as well.

PR #85, opened 2017-03-16
Resolved the issue of keyboard covering the textFields in different way.

hshrimali:
Attempted fixes for issue #21


src/main/res/layout/layout_authentication.xml
```diff
@@ -18,101 +17,119 @@
...
+            android:layout_height="8dp" />
+
+        <android.support.design.widget.TextInputLayout
+
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            android:theme="@style/Theme.AppCompat"
+            app:passwordToggleEnabled="true">
+
+            <EditText
```
csoni111: @hshrimali Why have you changed it to `EditText`? There's a slight difference between using `TextInputEditText` and `EditText`. Pls read this stackoverflow [answer](http://stackoverflow.com/a/36269036/6352482).

Anyway I have implemented this in another pr #90 resolving bug #87.

cpg:
Could you address the issues and resolve the conflicts at your convenience?

cpg:
The issues raised were not addressed. Too bad, because this is a useful feature. Closing.

hshrimali:
Actually I didn't understand what you wanted. @cpg

cpg:
Did you see the comments from @csoni111 ?

I got the impression you did not address those concerns for some reason.

hshrimali:
Okay so @csoni111,  @cpg ,  if I use textInputLayout in place of EditText it doesn't work fine with the setOnTouchListener and setOnFocusListener.  That's why I used it

csoni111:
@hshrimali you are already using the `TextInputLayout` in your code. I commented for using `TextInputEditText` inside the `TextInputLayout` instead of `EditText`. 

Also the `TextInputEditText` is extended from the `EditText` only. So every thing that `EditText` can do, automatically `TextInputEditText` can also do. Check here: 
[https://developer.android.com/reference/android/support/design/widget/TextInputEditText.html](https://developer.android.com/reference/android/support/design/widget/TextInputEditText.html)

hshrimali:
@csoni111 Inside the `TextInputLayout` , if I use `TextInputLayout` instead of `EditText`, then it shows null object reference in the code where I used `setOnTouchListener` whereas it works completely fine if I use `EditText`.

csoni111:
Why would one use `TextInputLayout` inside `TextInputLayout`? 

Use `TextInputEditText` inside `TextInputLayout`! That's what I have been saying since the first comment!. Please recheck the above comments.

hshrimali:
@csoni111 so sorry for not getting you correctly earlier. My bad. So, do I need to change the `EditText` into `TextInputEditText` and send one more PR because you said earlier that you have already done it in another PR ??

csoni111:
Check the code in the [current master](https://github.com/amahi/android/blob/master/src/main/res/layout/layout_authentication.xml#L48). It already contains the `TextInputEditText`. So you can make the changes (adding listeners and all) on the top of current master and then send the pr.

hshrimali:
Okay, done. I have sent the PR.

PR #114, opened 2017-03-19
Add IME_ACTION_GO listener on Password EditText in AuthenticationActivity

adithya321:
Clicking Go in keyboard after entering the password will call "SIGN IN" button's onClick listener.
Previously, it hid the keyboard and then the user had to click the "SIGN IN" button.

### DIFF
--- a/src/main/java/org/amahi/anywhere/activity/AuthenticationActivity.java
+++ b/src/main/java/org/amahi/anywhere/activity/AuthenticationActivity.java
@@ -30,7 +30,9 @@ import android.support.v4.content.ContextCompat;
 import android.text.Editable;
 import android.text.TextWatcher;
 import android.text.method.LinkMovementMethod;
+import android.view.KeyEvent;
 import android.view.View;
+import android.view.inputmethod.EditorInfo;
 import android.widget.EditText;
 import android.widget.TextView;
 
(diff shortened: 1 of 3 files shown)


cpg:
What exactly does this do?

What is pressing "Go"? Is it the enter/return key?

When I do that, the app vanishes and I see the main screen. I click on the windows button, find the app and I see it logged in.

I see errors, but I am nearly sure they are unrelated, no crash/exception.

Without this patch, it appears to do the same -- vanishing, except when I press return on the password field, the keyboard closes and I get to click on the Sign In button to make it login. At this point the app as a whole minimizes, so that may be a separate bug?

So, the question is -- why does this whole app minimization occur and should we proceed with that separately?

adithya321:
"Go" is the :arrow_right: key in Google Keyboard (it changes with keyboard).

potrait
landscape

I couldn't simulate the app minimization on pressing "Enter" key.

csoni111:
I think this may not be required because the user may want to see the password that he has entered, before actually sending the credentials. So according to me the previous functionality was good enough ie. on pressing `Enter` key in password field the keyboard minimizes, showing the SIGN IN button.

adithya321:
I expect to sign in after tapping `Enter` key in the password keyboard.
This removes an extra tap on the "SIGN IN" button.
He can still view the password after sending the credentials (if it was wrong).

PR #117, opened 2017-03-19
fix app closing on re-login issue

### LINKED ISSUE #116: App closes on re-login after logging out of the app
**Problem:** After logging out of the app, the user is redirected to login activity. Now if one tries to login again _(without closing the app)_ then on submitting valid id and password the authentication token is registered in the app but the **app gets closed**. This happens because of this change [here](https://github.com/amahi/android/pull/39/files#diff-215b51572e3fa3b8377310f3d528464bR171) which was added for resolving another bug.

csoni111:
Fixes #116


csoni111:
Previously `AuthenticationActivity` was being opened with NEW_TASK and CLEAR_TASK flags on clicking Logout button. Due to the flags the `NavigationActivity's` and `NavigationFragment's` running instances were destroyed which contained the listener for any `Account` updates. So on pressing the login button in `AuthenticationActivity`, it now closed the app since no listeners for `Account` changes are registered.

Now I have replaced the `AuthenticationActivity` with `NavigationActivity` in the intent to open activity on logout button press, so that the current running instance of `NavigationActivity` is brought to front and the activity is restarted, hence checking if the account exists or not. Then opening the `AuthenticationActivity`. This is the same process that is followed when the app is opened for the first time.

Also I have added a toast saying `Logged out successfully` to be displayed after logging out.

Issue #118, opened 2017-03-19
Double BackPress to exit the App.


doomers:
Almost all of the Apps are using Double BackPress to exit from the App.

cpg:
I think this is fluff. It's also bad UI, IMHO, for putting that in front of the user.

I see gmail doing an exit on one single tap. The app does that currently and the world does not crumble.

I will assign it for review, but I am not too inclined to do this unless you can offer evidence in the form of at least two or three major mainstream apps where I can see they do this. Or perhaps recent guidelines that ask app developers to do this.

doomers:
Okay @cpg have a look at these Popular Apps:
1.)Paytm

![paytm](https://cloud.githubusercontent.com/assets/11586392/24269300/c379b02e-1036-11e7-8019-442ffff1be38.gif)

2.)Freecharge

![freecharge](https://cloud.githubusercontent.com/assets/11586392/24269321/d4637e7e-1036-11e7-9bfc-497e4da56f8c.gif)

3.)Flipkart

![flipkart](https://cloud.githubusercontent.com/assets/11586392/24269362/f3454aca-1036-11e7-9e77-46565cda2a2b.gif)

cpg:
Ok, you proved some apps do it.

However, my design intuition says this is extra fluff that the user.

Fine, go ahead, unless @megabitdragon objects.

megabitdragon:
This is not needed.

PR #119, opened 2017-03-19
added double BackPress functionality

doomers:
BackPress Functionality added #118


src/main/java/org/amahi/anywhere/activity/NavigationActivity.java
```diff
@@ -337,4 +338,25 @@ protected void onPause() {
...
+
+    @Override
+    public void onBackPressed() {
+        if (doubleBackToExitPressedOnce) {
+            super.onBackPressed();
+            return;
+        }
+
+        this.doubleBackToExitPressedOnce = true;
+        Toast.makeText(this, "Please click BACK again to exit", Toast.LENGTH_SHORT).show();
```
csoni111: Please use the xml resource here that you have added in `strings.xml` instead of hardcoding.

doomers:
Thanks for pointing it out @csoni111.
Done

cpg:
Closing until we can review in #118 (which we will leave open) and we agree this is good UI.

PR #120, opened 2017-03-20
Added support for AppCompatActivity.

### LINKED ISSUE #96: Adding support for AppCompatActivity
For the time being the activities don't support AppCompatActivity. 

**Benefits of using AppCompatActivity**

It not only provide backwards compatibility to our app but also provides us the access to the Material Design Library too. After this we can easily add FloatingActionButtons, Snackbars, AlertDialogBoxes and other material design widgets easily which now we can't.

I'd be working on it.

octacode:
Fixes #96 
Changes Made :-
     1. Replaced the Activities with AppCompatActivity.
     2. Changed the parent theme to Theme.AppCompat.

### DIFF
--- a/src/main/java/org/amahi/anywhere/activity/NavigationActivity.java
+++ b/src/main/java/org/amahi/anywhere/activity/NavigationActivity.java
@@ -19,15 +19,13 @@
 
 package org.amahi.anywhere.activity;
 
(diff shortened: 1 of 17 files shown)


octacode:
The code is still buggy. Please don't merge it as it is. Will add the fix.

Issue #113, opened 2017-03-18
check whether the gmail application exist or not before launching


nikhilbalyan:
Problem Description:
Right now before launching the intent for gmail service we are not checking whether there is a mail application which can handle this intent or not. Like here while taking feedback

Expected Behavior:
We should add a check before launching the implicit intent whether there  is an application which can handle  the intent or not. if there is no activity that can handle that intent then showing the appropriate toast.
I would love to work on this issue.

PR #94, opened 2017-03-17
Tell a friend feature improved

doomers:
#92 Now Tell-a-friend Screen looks like this

screenshot_20170317-134544

PR #122, opened 2017-03-20
Add check whether an email application exist or not #113

nikhilbalyan:
Added check whether an email client exist or not before launching the implicit intent while taking feedback from settings.


src/main/java/org/amahi/anywhere/fragment/SettingsFragment.java
```diff
@@ -188,7 +189,12 @@ private void setUpApplicationVersion() {
 
 	private void setUpApplicationFeedback() {
 		Intent intent = Intents.Builder.with(getActivity()).buildFeedbackIntent();
-		startActivity(intent);
+		if (intent.resolveActivity(getActivity().getPackageManager()) != null) {
+			startActivity(intent);
+		}
+		else {
+			Toast.makeText(getActivity(), "NO Application Found", Toast.LENGTH_SHORT).show();
```
csoni111: Please declare the message in `Strings.xml` and use the xml resource here. This will help in translation process whenever it is done!

cpg:
Hi! Does this apply any more after we merged #94?

PR #135, opened 2017-03-21
Fix Icon changing to Image File Icon while scrolling

adithya321:


src/main/java/org/amahi/anywhere/adapter/FilesFilterBaseAdapter.java
```diff
@@ -197,7 +197,6 @@ void setUpImageIcon(ServerFile file, ImageView fileIconView) {
                 .centerCrop()
                 .fit()
                 .placeholder(getFileIcon(file))
-                .error(getFileIcon(file))
```
adithya321: getFileIcon(file) returns Image File icon (Mountain and Sun)

cpg:
Seems ok. However, `.pptx` has the right icon (like `.ppt`), however, `.xlsx` does not have the expected excel icon (`.xls`) and the `.docx` does not have the expected icon as `.doc`.

Is the icon determined by the extension or mime type?

I do see that the `.pptx` does have mime-type that mimics the ones in `.docx` and `.xlsx`, so I think this is not an issue from the server side.

adithya321:
The icon is determined by MIME type and I have added types for docx and xlsx.

PR #138, opened 2017-03-21
Attempted improvements for keyboard covering up login fields

hshrimali:
Attempted fix for #21 : keyboard covers up the textfields


cpg:
Lots [of errors in compilation](https://www.dropbox.com/s/utjgm0s01a0fvpi/Screen%20Shot%202017-03-21%20at%201.04.40%20PM.png?dl=0).

hshrimali:
But it's working completely fine on my pc.  Can you just tell me that what errors it is showing??

csoni111:
How is it possible? The code does not have proper imports in the `AuthenticationActivity.java` file. You are missing two import lines from the top: 
`import android.view.MotionEvent;`
`import android.view.inputmethod.InputMethodManager;`

Also even after adding these line, the code will build but you will get runtime error! As the ImageView that you have added to the code, you have never initialized its instance from the layout file. You need to add something like `(ImageView) findViewById(...);`.

I think if you are able to run the app because you might have made the changes to your local copy but forgot to push it to github. Otherwise its impossible to run the app with the same code that is present here!

src/main/java/org/amahi/anywhere/activity/AuthenticationActivity.java
```diff
@@ -51,6 +53,8 @@
 
 import javax.inject.Inject;
 
+import static org.amahi.anywhere.R.id.password_layout;
```
csoni111: This is unnecessary!

src/main/java/org/amahi/anywhere/activity/AuthenticationActivity.java
```diff
@@ -59,13 +63,47 @@
...
 	protected void onCreate(Bundle savedInstanceState) {
 		super.onCreate(savedInstanceState);
 		setContentView(R.layout.activity_authentication);
 
-		setUpInjections();
+    getUsernameEdit().setOnTouchListener(new View.OnTouchListener() {
+
+            @Override
+            public boolean onTouch(View view, MotionEvent motionEvent) {
+                getUsernameEdit().setOnFocusChangeListener(new View.OnFocusChangeListener() {
```
csoni111: Setting a listener inside another listener? seriously?
This will make a new copy of focus listener every time you touch the EditText. This can be simplified! Please revise your code.

src/main/java/org/amahi/anywhere/activity/AuthenticationActivity.java
```diff
@@ -59,13 +63,47 @@
...
+        });
+
+
+        getPasswordEdit().setOnTouchListener(new View.OnTouchListener() {
+
+            @Override
+            public boolean onTouch(View view, MotionEvent motionEvent) {
+                getPasswordEdit().setOnFocusChangeListener(new View.OnFocusChangeListener() {
+                    @Override
+                    public void onFocusChange(View view, boolean b) {
```
csoni111: Same Thing here too. 
Also one more thing you can do is that since the onFocusChangeListeners contain the same code for both the edit texts so you simply make a listener object once and pass it to both the edit texts.

src/main/java/org/amahi/anywhere/activity/AuthenticationActivity.java
```diff
@@ -59,13 +63,47 @@
...
-		setUpInjections();
+    getUsernameEdit().setOnTouchListener(new View.OnTouchListener() {
+
+            @Override
+            public boolean onTouch(View view, MotionEvent motionEvent) {
+                getUsernameEdit().setOnFocusChangeListener(new View.OnFocusChangeListener() {
+                    @Override
+                    public void onFocusChange(View view, boolean b) {
+                        if (b) {
+                            imageView.setImageDrawable(null);
```
csoni111: Instead of making the image drawable null you should set the VIsibility property of the ImageView. Set it to View.GONE when you need to hide the IV and to bring it back set it to View.VISIBLE.

src/main/java/org/amahi/anywhere/activity/AuthenticationActivity.java
```diff
@@ -153,161 +191,174 @@ private void setUpAuthenticationActionListener() {
...
+            imm.hideSoftInputFromWindow(getCurrentFocus().getWindowToken(), InputMethodManager.RESULT_UNCHANGED_SHOWN);
 
-			getUsernameEdit().addTextChangedListener(new TextWatcher() {
-				@Override
-				public void beforeTextChanged(CharSequence charSequence, int i, int i1, int i2) {
+            ViewDirector.of(this, R.id.animator_message).show(R.id.text_message_authentication_empty);
 
-				}
+            if (getUsername().trim().isEmpty())
+                getUsernameEdit().getBackground().setColorFilter(getResources().getColor(android.R.color.holo_red_light), PorterDuff.Mode.SRC_ATOP);
```
csoni111: On resolving the merge conflicts you have changed back the `ContextCompat.getColor` line to the `getResouces().getColor` which is deprecated and was previously resolved.

src/main/res/layout/layout_authentication.xml
```diff
@@ -18,111 +17,118 @@
...
-
-	</ViewAnimator>
-
-</LinearLayout>
+<ScrollView xmlns:android="http://schemas.android.com/apk/res/android"
+    android:layout_width="fill_parent"
+    android:layout_height="fill_parent">
+
+  <LinearLayout
+    xmlns:android="http://schemas.android.com/apk/res/android"
```
csoni111: This is unnecessary!

hshrimali:
Okay, I have implemented the things said by csoni111 in the new commits. Thank you @csoni111 for help. Now, you can review it. Thanks once again.

csoni111:
I was in the process of reviewing the code while you made the new commits. Please check the review comments once! @hshrimali

hshrimali:
Okay, @csoni111 , I got you. On it.

cpg:
could you resolve the conflicts, please?

csoni111:
@hshrimali I don't understand why you are setting the `onFocusChangeListener` inside `onTouchListener`? If you set the focus listener directly inside onCreate then also it will have te same effect.
Also there is no need for ScrollView, without it also it will have the same functionality.

hshrimali:
@csoni111 I accept your first point. But scroll view is needed otherwise keyboard will hide the "Sign In" button. And one more thing @cpg and csoni111 that last night something went wrong with my Android Studio. So, now I am sending a complete new and fresh PR for this issue. You can close this current one.

csoni111:
You can close this pr yourself. 
Also next time you send the pr, please do not make any code reformatting or indentation changes. This makes it difficult to read that in which lines you have made the changes as github will consider the whole file as changed due to indentation changes as in the case of this pr.

hshrimali:
Okay

Issue #98, opened 2017-03-17
Complete web page not showing


hshrimali:
After clicking on the "amahi anywhere" link in Overview.html, it shows only the half part of the web page (vertically)  instead of complete web page.

cpg:
Where is this link?

hshrimali:
Actually, there is a page that comes after clicking on Overview.html. On that page, there is a text "amahi anywhere" in blue color. When we tap on that text, it directs us to the this link (https://www.amahi.org/apps/amahi-anywhere). 
But only vertically half part of this page is displayed on the mobile screen that is the issue.

cpg:
If you do not specify where that `Overview.html` file is, or how a user of the app clicks on it, we can't do much about it.

hshrimali:
Okay, so after logging into the app , follow the following path.
Shares  >>  Introduction  >>  1 - Overview.html . After opening of Overview.html tap on the link "Amahi Anywhere" in the section "Install Amahi Anywhere in your HDA"

cpg:
Ok, I see that the web view is constraining the view port of the web page. I'm not sure if this is something that is fixable in the webview we use in the app or it's a result of the non-mobile friendly HTML/CSS in the amahi.org site. Pinch to reduce size does not seem to work.

If you can propose something to fix this, please suggest.

csoni111:
This can be done by adding a few settings to the web view. We need to enable Overview Mode and zoom controls on it. Can I send a pr for this fix?

Also on a separate note the amahi website must be made mobile-friendly!

cpg:
If you think it's a simple solution, we should fix it.

However, the apps area has to be tested as well if this settings affects other usage of the webview.

hshrimali:
@cpg I'm done fixing this issue, if csoni111 is not working on it. Can I send the PR ??

cpg:
Let's see your PR @hshrimali.

PR #141, opened 2017-03-22
Webview fixes. Attempt fixes for issue #98.

hshrimali:
Attempt fixes for issue #98


cpg:
fixes #98 (at least part of it)

PR #144, opened 2017-03-23
Attempted improvements for keyboard covering up login fields

hshrimali:
Attempted fixes for issue #21.

### DIFF
--- a/src/main/java/org/amahi/anywhere/activity/AuthenticationActivity.java
+++ b/src/main/java/org/amahi/anywhere/activity/AuthenticationActivity.java
@@ -23,6 +23,7 @@ import android.accounts.Account;
 import android.accounts.AccountAuthenticatorActivity;
 import android.accounts.AccountManager;
 import android.app.Activity;
+import android.content.Context;
 import android.graphics.PorterDuff;
 import android.os.Bundle;
 import android.support.design.widget.TextInputLayout;
@@ -33,7 +34,9 @@ import android.text.method.LinkMovementMethod;
 import android.view.KeyEvent;
 import android.view.View;
 import android.view.inputmethod.EditorInfo;
+import android.view.inputmethod.InputMethodManager;
 import android.widget.EditText;
+import android.widget.ImageView;
 import android.widget.TextView;
 
 import com.dd.processbutton.iml.ActionProcessButton;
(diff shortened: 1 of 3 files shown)


cpg:
Looking good. Will wait for a peer review.

cpg:
Looking good. Will wait for a peer review. Maybe @csoni111 and @doomers can take a look at the changes?

src/main/res/layout/layout_authentication.xml
```diff
@@ -26,9 +28,12 @@
 	android:paddingLeft="32dp"
 	android:paddingRight="32dp"
 	android:layout_width="match_parent"
-	android:layout_height="wrap_content">
+	android:layout_height="wrap_content"
+	android:descendantFocusability="beforeDescendants"
+	android:focusableInTouchMode="true">
 
 	<ImageView
+		android:id="@+id/logo"
```
csoni111: Please add this id to the landscape layout also! Currently the app is crashing in landscape mode!

src/main/java/org/amahi/anywhere/activity/AuthenticationActivity.java
```diff
@@ -203,6 +231,11 @@ public void afterTextChanged(Editable editable) {
 
 		}
 		else {
+            imageView.setVisibility(imageView.VISIBLE);
+
+            InputMethodManager imm = (InputMethodManager) getSystemService(Context.INPUT_METHOD_SERVICE);
+            imm.hideSoftInputFromWindow(getCurrentFocus().getWindowToken(), InputMethodManager.RESULT_UNCHANGED_SHOWN);
+
```
csoni111: If you are using these same three lines both in the `if` block and the `else` block. Then why not just add them outside the if / else , only once!

hshrimali:
@csoni111  Can I make these changes here only (i.e., on github) or need to do from terminal ??

doomers:
@hshrimali You can do it in the way you like, but as there are multiple changes, do it from terminal way,
as you can test the way that way, also if possible squash this commit with the previous ones, I personally like that way.

src/main/res/layout/layout_authentication.xml
```diff
@@ -17,7 +17,9 @@
   ~ You should have received a copy of the GNU General Public License
   ~ along with Amahi. If not, see <http ://www.gnu.org/licenses/>.
   -->
-
+<ScrollView xmlns:android="http://schemas.android.com/apk/res/android"
+	android:layout_width="fill_parent"
+	android:layout_height="fill_parent">
```
csoni111: Make both of these `match_parent` instead of `fill_parent`. As both are same but `fill_parent` is deprecated and was replaced by `match_parent` after api level 8.
Read [here](https://developer.android.com/reference/android/view/ViewGroup.LayoutParams.html#FILL_PARENT).

hshrimali:
Sorry for all these minute mistakes. Sending a new PR & closing this one.

csoni111:
No need to open another pr, you can make the changes in this one only. Either edit your previous commit or make another commit _(won't matter because finally it will be squashed & merged)_.

hshrimali:
Sorry @csoni111 , due to network problem , your comment came after closing the PR.

PR #145, opened 2017-03-23
Remove duplicate api.properties entry in .gitignore

adithya321:
`api.properties` entry was defined more than once.


cpg:
done separately.

PR #149, opened 2017-03-23
Attempted improvements for keyboard covering up login fields.

hshrimali:
Attempt fixes for issue #21 .

### DIFF
--- a/src/main/java/org/amahi/anywhere/activity/AuthenticationActivity.java
+++ b/src/main/java/org/amahi/anywhere/activity/AuthenticationActivity.java
@@ -23,7 +23,9 @@ import android.accounts.Account;
 import android.accounts.AccountAuthenticatorActivity;
 import android.accounts.AccountManager;
 import android.app.Activity;
+import android.content.Context;
 import android.graphics.PorterDuff;
+import android.media.Image;
 import android.os.Bundle;
 import android.support.design.widget.TextInputLayout;
 import android.support.v4.content.ContextCompat;
@@ -33,7 +35,9 @@ import android.text.method.LinkMovementMethod;
 import android.view.KeyEvent;
 import android.view.View;
 import android.view.inputmethod.EditorInfo;
+import android.view.inputmethod.InputMethodManager;
 import android.widget.EditText;
+import android.widget.ImageView;
 import android.widget.TextView;
 
 import com.dd.processbutton.iml.ActionProcessButton;
@@ -59,11 +63,17 @@ public class AuthenticationActivity extends AccountAuthenticatorActivity impleme
 {
 	@Inject
 	AmahiClient amahiClient;
+	ImageView imageView;
 
 	@Override
 	protected void onCreate(Bundle savedInstanceState) {
 		super.onCreate(savedInstanceState);
 		setContentView(R.layout.activity_authentication);
+
+		imageView = (ImageView) findViewById(R.id.logo);
+
+		FocusListener(getUsernameEdit());
+		FocusListener(getPasswordEdit());
 
 		setUpInjections();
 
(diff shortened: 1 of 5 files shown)


hshrimali:
@csoni111 Now you can review it.

doomers:
When you click on TextInputLayout(namely UserName,Password), You must be setting VIsibility of Amahi Icon to Gone.

But on backPress when keyBoard is on Screen Amahi Icon Visibilty remains to gone.
See Screenshot
screenshot_20170324-023403

hshrimali:
Okay so I think I need to override the method that gets fired when the back button is pressed and keyboard is open.  Right or any other suggestions.

doomers:
Yes, that should work portrait mode, but i doubt that would work for Landsape mode.
Still Go Ahead, make the changes and do add more commits to the PR squash them
as this number of commit doesn't looks good for a small issue.
Cheers

hshrimali:
Done with the issue that @doomers found out.

hshrimali:
@csoni111 can you please review it.

src/main/java/org/amahi/anywhere/activity/TestEditText.java
```diff
@@ -0,0 +1,47 @@
...
+import android.view.KeyEvent;
+import android.view.View;
+import android.view.inputmethod.InputMethodManager;
+import android.widget.EditText;
+import android.widget.ImageView;
+
+public class TestEditText extends TextInputEditText {
+
+    private static ImageView imageView;
+    private static EditText user, password;
```
csoni111: You should avoid making objects static like this. This will create memory leak as these objects would not be destroyed even when the AuthenticationActivity and its layout are destroyed (ie after login is successful). These will retain the space in memory and their reference to the specified objects. This must avoided at all costs! 

warning

Read [here](http://blog.nimbledroid.com/2016/05/23/memory-leaks.html) for more details on memory leak.

A better way to do this would be by using interfaces ie by making a custom listener in the TestEditText class and implementing it in the Activity. I hope you get this!

src/main/java/org/amahi/anywhere/activity/TestEditText.java
```diff
@@ -0,0 +1,47 @@
...
+import android.content.Context;
+import android.support.design.widget.TextInputEditText;
+import android.util.AttributeSet;
+import android.view.KeyEvent;
+import android.view.View;
+import android.view.inputmethod.InputMethodManager;
+import android.widget.EditText;
+import android.widget.ImageView;
+
+public class TestEditText extends TextInputEditText {
```
csoni111: Please give it a proper name! 
`TestEditText` does not really signify anything.

src/main/java/org/amahi/anywhere/activity/AuthenticationActivity.java
```diff
@@ -23,7 +23,9 @@
 import android.accounts.AccountAuthenticatorActivity;
 import android.accounts.AccountManager;
 import android.app.Activity;
+import android.content.Context;
 import android.graphics.PorterDuff;
+import android.media.Image;
 import android.os.Bundle;
 import android.support.design.widget.TextInputLayout;
```
csoni111: Please always clean the unused dependencies. Here two of them are unused!
`import android.media.Image;`
`import android.support.design.widget.TextInputLayout;`

src/main/java/org/amahi/anywhere/activity/TestEditText.java
hshrimali: Sorry, but I didn't get this..

cpg:
Ok, what if we just abandon this and just use #176 ?
It's a one-line change and seems to work well.

cpg:
Good work, @hshrimali.

Issue #42, opened 2017-03-08
Issues with HomeUp button


octacode:
When inside a directory Home up button directly takes the user to the list of shares instead of taking it in the upper folder.
For Example :- if we are in Demo/music and if home Up is pressed. We'll be redirected to the Shares fragment instead of Demo.
I am sending a PR for this.
Thanks

cpg:
I remember there was some discussion about this when we created the app and I think it was determined that navigation back was to be done with the proper "back" button, and the home button was to be to take the user to the top level of the app navigation.

I would like to seek the advice of people with expertise of what matters with the latest best practices on this. @megabitdragon may have a good opinion.

adithya321:
We can have a Home icon to take the user to the top level like [Amaze File Manager](https://play.google.com/store/apps/details?id=com.amaze.filemanager).

<img width="515" alt="amaze" src="https://cloud.githubusercontent.com/assets/3854934/24161480/87809ea0-0e8a-11e7-8cfb-38a54edba0f8.png">

octacode:
@cpg you views on this??

cpg:
Based on [the guidelines](https://material.io/guidelines/patterns/navigation.html#navigation-usage), it seems like

- the up button (at the top, inside the app header), should be hierarchical. In other words, take the user one level "up" in the hierarchy within the app. We are not doing this at the moment.
- the back button (at the bottom, possibly a hardware button) should be chronological. I have heard this is kept automatically. I am not sure if we do anything in the app about this. We probably should *not* if it's kept automatically?

Note: there is the subtle question about the meaning of "screen" in the guidelines. When the user is navigating in our app -- is she presented with a new "screen" when they enter a share and a folder or more inside a share and a folder? Or is she in the same "screen" with different content?

My feeling is that each share and folder should be considered a "screen" for our purposes.

What does #43 do in this respect, and does it match the guidelines and our interpretation of them in my comments above?

octacode:
When the user navigates from Share to a folder that's when a new screen is presented. However, When the user navigates from folder to another folder(nested) then the screen remains same but the content is updated.
I too feel we should consider each folder and share as a new screen.
If we consider each folder and share as a new screen then #43 matches the guidelines.

cpg:
I see the back button does not seem to be chronological at the moment.

Does the app control it?

If so, can we just not do anything and get it working chronologically as it's meant to be?

octacode:
[...]

PR #157, opened 2017-03-27
Fix fragment transaction on home back button click in the file explorer

guenodz:
When a user click the back home button in ServerFilesActivity, the app will go directly to NavigationActivity. However, this also happens when a user want to back to the parent folder. The changes I made here solved the problem and the app works fine.
(the other highlighted files are not modified, it's just a code cleaning/formatting)


adithya321:
There is an ongoing discussion about this in #42. Also, do not submit changes that include reflow/reindent with code changes. Submit those separately.

octacode:
can close this one...As there's already a PR regarding this. #43

guenodz:
Sorry, I didn't see the issue #42 and PR #43.

Issue #158, opened 2017-03-28
Reformat Code


adithya321:
I am used to reformatting code (Cmd+Alt+L in mac) and I have to undo that everytime as it changes a lot in every file. For example, PR #108 and #157. 

I use the Default code style scheme in Android Studio but we can use any code style by committing `.idea/codeStyleSettings.xml`. By doing this everyone can reformat code to match the style before committing. 

@cpg How do you think we can reformat the whole codebase?

guenodz:
Personally I use the official Google code style for Android projects [here](https://github.com/android/platform_development/blob/master/ide/intellij/codestyles/AndroidStyle.xml). We can use it or use another common style, or we can use the default code style in IntellijIdea. What do you think ?

cpg:
We discussed about this and agreed to adopt this. The only remaining thing is to add a section (before the `Generating the documentation` section) in the README titled `Code Style Convention` with details for other developers to follow when thy contribute in the future.

After you do that (separate PR) we can close this. Thanks!

adithya321:
I am working on it and I will send a PR soon.

PR #159, opened 2017-03-28
Reformat Code

adithya321:
Per suggestion from #158 
Ran `Reformat Code` with `Optimize imports` and `Rearrange entries` options checked.
I use the `Default` scheme in Code Style.

screen shot 2017-03-28 at 05 48 48

### DIFF
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -48,8 +48,7 @@
         <activity android:name=".activity.ServerFilesActivity">
             <meta-data
                 android:name="android.app.searchable"
-                android:resource="@xml/searchable">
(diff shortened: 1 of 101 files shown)


guenodz:
There are conflicts due to the use of incompatible layout propriety (android:foreground requires a min api of 21) in these layout files.

doomers:
@adithya321 maybe they are using some other code styles other than the default android studio one,  as most of developers goes with the style they like.

adithya321:
@doomers Currently, the code style is not the same across all files and this can be avoided if a code style is shared. I'm suggesting that we use the `Default` Android Studio code style as everyone can follow it.

Issue #153, opened 2017-03-23
Test Cases For AndroidManifest permissions


nikhilbalyan:
Problem Description:
Right now we do not have the test cases for Permissions in AndroidManifest.xml and if someone delete them then that could be problematic.

Expected Behavior:
Adding the Unit Test to the amahi for android manifest permissions.

PR #166, opened 2017-03-28
Add AndroidManifest Permissions Test

nikhilbalyan:
Added test cases to check verify that all the AndroidManifest Permissions exist and none of them gets deleted by mistake issue #153

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -124,6 +124,9 @@ dependencies {
     }
     provided "com.squareup.dagger:dagger-compiler:1.2.5"
 
+    testCompile 'org.robolectric:robolectric:3.1.2'
+    testCompile 'junit:junit:4.12'
+    testCompile 'org.robolectric:shadows-multidex:3.0'
     androidTestCompile 'com.android.support.test.espresso:espresso-core:2.2.2'
     androidTestCompile 'com.android.support.test:runner:0.5'
     androidTestCompile 'com.android.support:support-annotations:25.3.1'
--- a/src/test/java/org/amahi/anywhere/PermissionTest.java
+++ b/src/test/java/org/amahi/anywhere/PermissionTest.java
@@ -1,0 +1,61 @@
+package org.amahi.anywhere;
+
+import org.hamcrest.Description;
+import org.hamcrest.StringDescription;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.robolectric.RobolectricTestRunner;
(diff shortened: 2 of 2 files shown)


cpg:
Could you explain what kind of testing is this doing?
- When do these tests run? by travis at compile time or dynamically while the app runs?
- How/when is the error detected?
- How/when is the error presented? to the developer? or the end-user?

I saw you had another similar PR and we have not set up a testing strategy for the app, so I would like to know more about this, thanks!

nikhilbalyan:
- When do these tests run? by travis at compile time or dynamically while the app runs?
@cpg we need to manually run these tests by rightclicking on the PermissionsTest and then clicking run and in subsequent runs we can run these tests by selecting from the run configuration. (Note this is from Android Studio IDE Perspective). This is an automated test but we need to initiate them.

- How/when is the error detected?
The test runs successfully as far as non of the AndroidManifest permissions is missing but will fail if any of the AndroidManifest Permissions goes missing.

-How/when is the error presented? to the developer? or the end-user?
The Result/error will be presented to the developer in Run console of Android Studio.

Categorization of this test:
This test is categorized under test category hence i have put this in the directory **module-name/src/test/java/.**.  And i am using the Android Junit and Android Roboelectric for testing and moreover i have specified the default sdk as 23 since Roboelectric does not support the latest sdk 25. 

**Roboelectric** is useful to inject the Android Sdk based Dependencies
**JUnit** is kind of defacto tool which is used for android unit testing 

Testing Strategy for Amahi:
Initially we need to chose an architecture MVP(Model View Presenter)/ MVC(Model View Controller) architecture. Then after identifying either of the architecture we should go ahead with choosing the tools for testing.

**Android Test(Instrumented Test):-** These tests are used for testing user interaction with User Interface. for Instrumentation tests we can use tools like Espresso

**Test(JVM Test):-** These tests run locally on JVM and usually runs very quickly. For them we can use tools like Junit which is used for Unit testing.

**Dependency Injection:-** Further we should chose a dependency injection framework like dagger 2 for injecting dependencies and removing the flakiness of test cases. (they usually helps to run test in isolation)

Furthermore there is a lot of scope in testing in Android and i am pretty sure that a lot of you will be having great ideas about which strategy to use.


Earlier i send a PR which was for Android Runtime Permissions i haven't yet send an PR like adding test for AndroidManifest those PR were totally different and were supposed to solve entirely different issues

Issue #136, opened 2017-03-21
Application crashed on clicking the next song icon


nikhilbalyan:
Problem description:
app crash when the user play the next song.

Steps to reproduce:
step:-1 login to application
step:-2 click share menu item from the navigation bar
step:-3 click the demo
step:-4 click the music
step:-5 click the 02 Meghan Trainor song
step:-6 click play the next song icon

And the application crash as soon as user click the next song icon or the previous song icon
I would love to work on this issue

cpg:
I can reproduce this.

harshvardhanpro:
@nikhilbalyan are you working on this issue ?

joydeepsaha05:
@cpg This can be closed.

PR #170, opened 2017-03-30
fix crash on clicking next song

harshvardhanpro:
solved issue #136 but now there is new issue. Next button is working but now the progress bar and current duration is not working.

### DIFF
--- a/src/main/java/org/amahi/anywhere/activity/ServerFileAudioActivity.java
+++ b/src/main/java/org/amahi/anywhere/activity/ServerFileAudioActivity.java
@@ -515,19 +515,19 @@ public class ServerFileAudioActivity extends AppCompatActivity implements Servic
 		stopService(intent);
 	}
 
-	private static final class AudioControlsNextListener implements View.OnClickListener
+	private final class AudioControlsNextListener implements View.OnClickListener
 	{
 		@Override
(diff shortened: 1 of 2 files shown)


cpg:
please file a separate issue (properly titled) for that other problem you are facing now.

csoni111:
@harshvardhanpro I don't think so you have resolved the problem! You have suppressed the problem by removing some functionality. The Bus event methods were meant for purpose, removing those does not solve this! Infact you have removed the calls to line [#291](https://github.com/amahi/android/blob/master/src/main/java/org/amahi/anywhere/activity/ServerFileAudioActivity.java#L291) and [#299](https://github.com/amahi/android/blob/master/src/main/java/org/amahi/anywhere/activity/ServerFileAudioActivity.java#L299) in the `ServerFileAudioActivity.java`. 

The actual error was a simple one and that was because of line [#315](https://github.com/amahi/android/blob/master/src/main/java/org/amahi/anywhere/activity/ServerFileAudioActivity.java#L315). In this we are still calling the old `getActionBar` method (which returns a NPE) while we should have called `getSupportActionBar` method. This must not have been updated while we were updating from Activity to AppCompatActivity.

Please correct this and **undo** the changes you have done in this pr!

Issue #179, opened 2017-03-31
accent colour not following material design guidelines


doomers:
after #175 ,with the addition of primary and accent color, app is looking nicer than before
but if we go by material design guidelines accent color should be a light colour of 200
https://material.io/guidelines/style/color.html#color-color-palette

but we are using orange of value 500, which looks quite  odd, 
and also as our app has a dark theme, we can use orange colour of value 300 also and it is also looking nicer than the 200 one.

cpg:
done

PR #180, opened 2017-03-31
changed value of accent color from 500 to 300

doomers:
#179 done.


cpg:
looks like there was a clash

cpg:
produced a conflict. no clear resolution. closing for now

Issue #148, opened 2017-03-23
Replace Toasts with Snackbars.


octacode:
#142 

I will replace the given Toasts with Snackbars.

nikhilbalyan:
@octacode please explain the specific places where you want to replace Toasts with Snackbars because both of them have their own use cases you can't just randomly go and replace all Toasts with snackbars

ak2711:
can i work on this issue?

cpg:
@ak2711 please let @octacode confirm he's working on it.

ak2711:
Are there any issues that are not assigned

cpg:
You can go through the issues and see if someone said they will work on it. You could also simply add new features you would like to see in the app after using it. Or make fixes/improvements that are sensible to you.

octacode:
Sorry I got up in submissions will update you all with the PR status very soon.

octacode:
**Settings Fragment**:-
While Signing out - Toast is fine as there's an activity change and we want to display the message over the change which isn't possible with Snackbars because Snackbar's instance dies with the activity which isn't the case with Toast.
In setUpApplicationFeedback() and setUpApplicationRating() :- Snackbar will look awesome. :)
**ServerFilesFragment**:- Here replacing Toasts with Snackbars isn't going to do us any harm.
**GooglePlaySearchFragment**:-Here too we can replace the Toasts with Snackbars.

Thanks

PR #181, opened 2017-03-31
Replaced the selected Toasts with Snackbars.

octacode:
I have replaced the selected Toasts with Snackbars and localised the hardcoded Permission granted and permission denied Strings.
Please look #148 
Thanks

### DIFF
--- a/src/main/java/org/amahi/anywhere/fragment/GooglePlaySearchFragment.java
+++ b/src/main/java/org/amahi/anywhere/fragment/GooglePlaySearchFragment.java
@@ -19,13 +19,13 @@
 
 package org.amahi.anywhere.fragment;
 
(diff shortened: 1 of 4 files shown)


octacode:
I've added an action to the permission snackbar which launches an intent to the Amahi app in the settings. This snackbar will only be displayed when user denies the permission to share.
screenshot_20170402-125551

cpg:
The sentence is not grammatically correct. I suggest it should be something like this:

`You have denied the permissions for XYZ. Enable permissions to continue.`

Then `Permission` should probably be `Permissions`.

I'm not a heavy Android user, but perhaps this is a common thing and we can see what other *popular* apps say?

octacode:
Alright I'll fix it.

PR #187, opened 2017-04-04
Audio player updates

### LINKED ISSUE #155: Replacing deprecated RemoteControlClient with MediaSession's Media Controls
In the [AudioService.java](https://github.com/amahi/android/blob/master/src/main/java/org/amahi/anywhere/service/AudioService.java#L89) we are using the old and deprecated `RemoteControlClient` _( deprecated since api 21 or android 5.0)_ to play the music in background service. 
This needs to be replaced by `MediaSessionCompat`.
You may check the changes here on [developer.android.com](https://developer.android.com/about/versions/android-5.0-changes.html#BehaviorMediaControl).

The replacement will offer greater control on the music player and allow music controls to be shown on LockScreen. I shall be working on this issue over the next week.

csoni111:
Resolves #155 

This pr implements following:
- Replaces deprecated `RemoteControlClient` with `MediaSession's Media Controls`
- Adds play/pause, next, previous buttons to notification as well as lock screen
- Adds Album Art to the lockscreen background

lockscreen

### DIFF
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -30,6 +30,7 @@
     <uses-permission android:name="android.permission.MANAGE_ACCOUNTS" />
     <uses-permission android:name="android.permission.USE_CREDENTIALS" />
     <uses-permission android:name="android.permission.WAKE_LOCK" />
+    <permission android:name="android.permission.MEDIA_CONTENT_CONTROL" />
 
     <application
         android:name=".AmahiApplication"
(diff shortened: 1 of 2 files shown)


cpg:
I think this may be a candidate to having introduced a big volume of crash reports. See #198. Only appears to fail in local shares, not in the cloud demo server (strange, but hey, it's not a real HDA).

@csoni111 ?

csoni111:
Will look into it and send a fix asap.

PR #193, opened 2017-04-11
Replace webviews with custom tabs, so that htpassword and other things work better

### LINKED ISSUE #186: Option to resend request if failed
Whenever there is a network error, add an option to tap the error to retry. This would be better from the UX perspective.

### LINKED ISSUE #192: bug: webapps tab fails if the app has htpasswd access set
you get automatic failure instead of asking for username/password when having htpasswd enabled on a webapp.

octacode:
Replaced the web view with custom Tabs.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -45,7 +45,7 @@ android {
         versionName buildVersionName(versionMajor, versionMinor, versionPatch)
         versionCode buildVersionCode(versionMajor, versionMinor, versionPatch)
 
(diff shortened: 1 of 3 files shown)


joydeepsaha05:
@octacode I don't see how this pull request fixes #186 
I've opened another pull request #194 to fix that issue.

octacode:
My bad! It's #171

PR #194, opened 2017-04-11
[Do Not Merge] Added "Tap to retry" option whenever server request fails

### LINKED ISSUE #186: Option to resend request if failed

joydeepsaha05:
Fixes #186


cpg:
Reopen when ready

PR #199, opened 2017-04-13
Fix crash when playing songs

### LINKED ISSUE #198: lots of crashes being reported by users playing music
W're getting lots of crashes in ServerFile.getMime calls. To reproduce, click on a music file to play! Play starts, but it crashes quickly and system asks the user to report a crash report. Music continues playing, however.

This appears to occur only on local share files. It appears to work from the `Welcome to Amahi` demo cloud server.

we're getting lots of crash reports that look like this. Any idea what could be causing this? and what should be a reliable fix?

```
Fatal Exception: java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.String org.amahi.anywhere.server.model.ServerFile.getMime()' on a null object reference
       at org.amahi.anywhere.util.Intents$Builder.getServerFileActivity(Intents.java:99)
       at org.amahi.anywhere.util.Intents$Builder.buildServerFileIntent(Intents.java:121)
       at org.amahi.anywhere.service.AudioService.setUpAudioPlayerNotification(AudioService.java:289)
       at org.amahi.anywhere.service.AudioService.playAudio(AudioService.java:389)
       at org.amahi.anywhere.service.AudioService.handleAudioFocusGain(AudioService.java:477)
       at org.amahi.anywhere.service.AudioService.onAudioFocusChange(AudioService.java:455)
       at android.media.AudioManager$FocusEventHandlerDelegate$1.handleMessage(AudioManager.java:3119)
       at android.os.Handler.dispatchMessage(Handler.java:102)
       at android.os.Looper.loop(Looper.java:158)
       at android.app.ActivityThread.main(ActivityThread.java:7237)
```

joydeepsaha05:
Fixes #198

### DIFF
--- a/src/main/java/org/amahi/anywhere/service/AudioService.java
+++ b/src/main/java/org/amahi/anywhere/service/AudioService.java
@@ -386,13 +386,17 @@ public class AudioService extends MediaBrowserServiceCompat implements AudioMana
 	public void playAudio() {
 		audioPlayer.start();
 		setMediaPlaybackState(PlaybackStateCompat.STATE_PLAYING);
-        setUpAudioPlayerNotification(audioMetadataFormatter, audioAlbumArt);
+		if (audioMetadataFormatter != null) {
+			setUpAudioPlayerNotification(audioMetadataFormatter, audioAlbumArt);
+		}
 	}
 
 	public void pauseAudio() {
 		audioPlayer.pause();
 		setMediaPlaybackState(PlaybackStateCompat.STATE_PAUSED);
-        setUpAudioPlayerNotification(audioMetadataFormatter, audioAlbumArt);
+		if (audioMetadataFormatter != null) {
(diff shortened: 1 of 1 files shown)


cpg:
@octacode @csoni111 How is this looking?

csoni111:
This does not end the problem, this only fixes things temporarily. 
I figured out that initially in the pr #187, I called the [setUpAudioPlayerNotification()](https://github.com/amahi/android/pull/199/files#diff-73396d7ae4339ff7848b0e2e8d1cfdf1R390) from inside the play/pause functions to update the play/pause icon on the notification. But that is not the correct way to do it. Instead the MediaSession allows to set a `media session token` on the notification itself so that it can be automatically updated as the media state is changed.

octacode:
I tested it. However it temporarily solves our issue but I am still getting the "Amahi has stopped working" error messages after every phone call or when the device is idle. Looking into the reason for the crash.

cpg:
Fixed by #201 in a seemingly more general way

PR #210, opened 2017-04-20
Replaced the ListView with the recycler view in the NavigationDrawer.

octacode:


cpg:
What does this actually do?

octacode:
As after android L recycler view was introduced and it's scrolling performance is better than ListView. This PR replaces the list view present in the Navigation Drawer with Recycler View.

PR #211, opened 2017-04-29
Video Landscape and portrait#76 issue

RanchoDbest:
- Added rotation button activity_server_file_video.xml
- On rotation button, I have set the code for landscape and portrait screen.


src/main/res/layout/activity_server_file_video.xml
```diff
@@ -31,17 +31,28 @@
...
             android:layout_width="match_parent"
             android:layout_height="match_parent"
             android:layout_gravity="center"
             android:fitsSystemWindows="false" />
+        <Button
+            android:id="@+id/button"
+            android:layout_width="20dp"
+            android:layout_height="20dp"
+            android:background="@drawable/rotation"
+            android:layout_marginTop="80dp"
```
csoni111: This top margin of 80dp looks odd. According to me, it should be something like this: the top margin can be 20dp initially and the button should slide up and down as the controls are shown or hidden. This may be done by some use case of `android:fitsSystemWindows="true"`. Please check if you can make it work like this.

src/main/res/layout/activity_server_file_video.xml
```diff
@@ -31,17 +31,28 @@
...
             android:id="@+id/surface"
             android:layout_width="match_parent"
             android:layout_height="match_parent"
             android:layout_gravity="center"
             android:fitsSystemWindows="false" />
+        <Button
+            android:id="@+id/button"
+            android:layout_width="20dp"
+            android:layout_height="20dp"
+            android:background="@drawable/rotation"
```
csoni111: `src/main/res/drawable-xxhdpi/exit_screen.png` has not been used anywhere in the code. Seems like it's extra and can be removed!

cpg:
@RanchoDbest ... we have not heard from you after the feedback provided.

Ultimately, there is one question: why? why do this feature? are there people who need to watch a movie in portrait mode?

I have not checked the UI, however, if it's very *very* unobtrusive, maybe we can merge this one in the remote hopes that it may help someone somewhere one day.

cpg:
No updates. Closing.

Issue #69, opened 2017-03-10
Swipe to change song.


Ishmeetsingh97:
I want to implement a feature which changes the song on swiping the AudioActivity just like done in Google Play Music. Can i work on it ?

cpg:
This is a useful idea.

Ishmeetsingh97:
Sure. I will send a PR soon.

doomers:
@cpg can i work on this issue?

csoni111:
As discussed in pr #212 following features need to be implemented:

- Pre fetching metadata for songs
- Implementing proper caching mechanism for song metadata (along with album art bitmap)
- Replacing Album art ImageView with ViewPager
- Thumbnail album art previews in place of icons for audio files

Yasha08:
Hi, what is this app about? I'm a UX/UI designer and is looking to contribute. Is this a music player app?

cpg:
[Read all about it](https://www.amahi.org/android), not surprisingly, with a quick search.

ArihantJain456:
I would like to work on this.

mohitsvnit:
@ArihantJain456 are you working on this?

ArihantJain456:
ye

mohitsvnit:
Because I already completed half of it.

ArihantJain456:
I have also completed most of it, only linking metadata with viewpager is left.
you should ask first before working on it.

mohitsvnit:
OK no problem. Actually I started working on it 3 days ago and that time nobody was working on it.

ArihantJain456:
Then why you haven't comment?

mohitsvnit:
Actually I thought no body is working on it. I didn't got notification that you commented. Sorry.

cpg:
It's our fault for not confirming @ArihantJain456 as assigned to it. Too many things going on and it fll through the cracks. Please try and be cooperative. It's not a competition.

PR #212, opened 2017-05-02
Add swipe to change songs

joydeepsaha05:
Implements #69

### DIFF
--- a/src/main/java/org/amahi/anywhere/activity/ServerFileAudioActivity.java
+++ b/src/main/java/org/amahi/anywhere/activity/ServerFileAudioActivity.java
@@ -51,6 +51,7 @@ import org.amahi.anywhere.server.model.ServerShare;
 import org.amahi.anywhere.service.AudioService;
 import org.amahi.anywhere.util.AudioMetadataFormatter;
 import org.amahi.anywhere.util.Intents;
+import org.amahi.anywhere.util.OnSwipeListener;
 import org.amahi.anywhere.util.ViewDirector;
 import org.amahi.anywhere.view.MediaControls;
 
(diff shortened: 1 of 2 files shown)


csoni111:
@joydeepsaha05 the code looks clean and it's functioning properly. 
But I think it would be better if while swiping, the song art slides along with the swipe and next/previous song art becomes visible _(just as it happens in Google Play Music)_. This way it will give user a visual feedback that on swiping, the song can be changed!

**But a lot of changes would be required for doing this:**

- Instead of a single ImageView, a ViewPager would be required. (Here implementation from `ServerFileImageActivity` can be reused.)
- Instead of fetching metadata for single audio file, metadata would be fetched for all the files at once (or we can limit it to several files which are in the vicinity of the current file).
- If this is done then the code in this pr will become useless.

**How will it benefit?**

- No delay would be there in starting a song as the metadata would already have been fetched.
- We would be able to display the Song Artwork in the list of songs as their icons (just as it happens in the image previews).
- Changing to Next/Previous songs would be quick and without delay.
- And of course there will be visual feedback while swiping the songs.

cpg:
If we were to implement this in two steps, would this be a first step, then add the rest of things @csoni111 suggested?

Or is it better/easier/faster to just do the (admittedly, quite nicer) suggestions all in one shot?

joydeepsaha05:
@csoni111 I agree that having visual feedback while swiping would be much better, but there will be a lag since the songs are not stored locally. Having the songs stored in memory would leave a huge footprint. What do you think?

@cpg The code in this PR would be useless if we were to do the above changes.

csoni111:
@joydeepsaha05 We do not have to store the songs locally, we just have to store the fetched metadata of those songs in Cache (Either memory based or disk based). This can be done by using one of the cache management libraries in android: [dualcache](https://github.com/vincentbrison/dualcache), [Reservoir](https://github.com/anupcowkur/Reservoir) etc. 
Also you can have a look at this music player's [source code](https://github.com/sdsmdg/MusicDNA/tree/master/app/src/main/java/com/sdsmdg/harjot/MusicDNA/imageLoader). This may be useful.

joydeepsaha05:
@csoni111 I'll look into that. Closing this PR.

PR #216, opened 2017-06-05
Video player updates

csoni111:
This pr implements following:
- Adds Subtitle View in the vlc video player (detects a subtitle with same name as the video file)
- Adds buffering indication in the seek view
- Some minor bugs resolved


cpg:
I think we should consider reverting this PR and sticking it to a branch until such times that the subtitles are not hardcoded on.

Any objections?

We also need (though at much lower priority) to add the [new mime types and remove extension checks per this bug #2235](https://bugs.amahi.org/issues/2235).

PR #220, opened 2017-06-09
Login, Main and Introduction activities for Android TV

octacode:
Added fragment callbacks to the Navigation Activity.

Implemented the shares and apps with the mainTVfragment.

Implemented fragment callbacks for Servers.

Implemented the Introduction Activity which will only be launched on first run.

Fixed the dual-log in bug.

Fixed the mobile app crash.

Moved the presenters from mainTVfragment to a diff package

Added logo to the Apps in the mainTVFragment.


src/main/java/org/amahi/anywhere/activity/NavigationActivity.java
```diff
@@ -78,6 +93,25 @@ protected void onCreate(Bundle savedInstanceState) {
 		super.onCreate(savedInstanceState);
 		setContentView(R.layout.activity_navigation);
 
+		if(isATV()){
+			Boolean isFirstRun = getSharedPreferences("PREFERENCE", MODE_PRIVATE).getBoolean("isFirstRun", true);
```
csoni111: There's a separate class [util/Preferences.java](https://github.com/amahi/android/blob/master/src/main/java/org/amahi/anywhere/util/Preferences.java#L55) in our project. We could use that for accessing shared preferences instead of polluting the code here.

Issue #239, opened 2017-07-10
[ATV] when selecting server, clicking on "go back" makes the app spin


cpg:
one can leave the server select area with the back button, however, selecting the "go back" option (which should also take one back, like the back button), yields a spinner.

the app is not hung, however. can we just remove this option if the back button works?

octacode:
#241 fixes this issue.

PR #241, opened 2017-07-10
Fixed the server select bug by removing the back option.

octacode:
Removed the back press option from the server selection fragment.


cpg:
Why not fix it?

Is having no back option the usual in android?

If removing, I would have also removed the other back option from the option of connection selector.

cpg:
fixes #239

Issue #243, opened 2017-07-12
Auto Upload photos


megabitdragon:
A option to auto-upload pictures from camera should be added. This should be enabled by a switch in the settings menu. Once enabled, two more field should show up specifically, path to the upload folder and a switch that will indicate if  pictures should be uploaded on cellular data or not. The images need to be added to a queue and uploaded only on WiFi if the cellular data is disabled. Otherwise the images will be uploaded automatically.

csoni111:
When will the auto upload trigger? Does this mean that as soon as the user takes a new photo from any of the camera application available on his device, the upload shall trigger or the photo will have to taken from inside amahi app?

megabitdragon:
Is should work for every camera app or at least for the default one. I think it should trigger as soon as a photo is taken. We might put a delay on place just in case the person decides to delete the picture immediately after is taken. Maybe add the image to a queue, and after a determined delay, if picture is still in the folder upload it.

csoni111:
I see two options for achieving this:

1) We can use [FileObserver](https://developer.android.com/reference/android/os/FileObserver.html) to observe a particular directory for any new file addition (recursively inside the directory listing). Then on callback, we start a background service which uploads the file to the server.

2) Another option is to register a `ACTION_MEDIA_SCANNER_SCAN_FILE` Broadcast Receiver. This [question](https://stackoverflow.com/questions/21311207/how-to-get-notified-of-each-new-image-being-visible-to-the-gallery-app) can be referred for more info on this.

3) We can also listen to [Camera#ACTION_NEW_PICTURE](https://developer.android.com/reference/android/hardware/Camera.html#ACTION_NEW_PICTURE) brodcast but this may result in not getting any callbacks from non-default camera apps or if a screenshot is taken.

Having a `FileObserver` will allow us to extend this feature for any file and not just image/media files. Also I read [somewhere](https://github.com/owncloud/android/issues/6) that some devices do not send the MediaScanner intent properly so we might miss out on some devices.

[...]

PR #253, opened 2017-07-25
Auto upload feature

csoni111:
Implements the auto/instant upload feature for images as soon as they are captured using camera. (#243)

Adds following things:

1) **Upload Manager** : For managing the uploads and processing upload queue one by one.

2) **Upload Service** : The background service which pushes urls of images to be uploaded to the sqlite db, initiates the upload manager and handle notifications for upload start, progress, success and error.

3) **Upload Settings** : The preference fragment which shows the UI for enabling/disabling auto upload and selecting the default upload HDA, Share and Path.

4) **Camera Receiver** : The Broadcast Receiver which listens for NEW_IMAGE broadcast sent by android device.

5) **Upload Queue Db** : The SQLite database for maintaining the upload queue in a persistent storage. 

6) **Upload Queue Db Helper** : For accessing the SQLite database. Contains functions for CRUD operations on the db.

7) **PhotosContentJob** : Job to monitor for changes to photos in the media provider.

8) **NetConnectivityJob** : Job to start upload service on gaining network connectivity.


csoni111:
The current approach followed for implementing the auto-upload feature is something like this:

**On android 7 and above:**
There are two jobs ->

- *PhotosContentJob* : It monitors content uri of image directory, and on any change, it starts a background service (*UploadService*) with image urls passed to the service which then adds those to the sqlite db, then checks if the device is connected to internet (also if the user has disabled "Allow on Cellular Data" in settings then checks for wifi connectivity), and starts the uploading process if connected.
It is scheduled once using JobScheduler at the start of the application, then re-scheduled after every subsequent run of this job.

- *NetConnectivityJob* : It starts when the device is connected to the internet and starts the same service as above but without any image urls passed to it.
It is also registered once at the start of the application. It is then re-scheduled from inside the *UploadService*, if it is detected that the network conditions are not met.

**On below android 7:**
Instead of the two jobs, there are two broadcast receivers ->

- *NEW_PICTURE* : It triggers whenever any image is captured on the device and does the same as done by *PhotosContentJob*.  
- *CONNECTIVITY_CHANGE* : It triggers on any connectivity change in the device and starts *UploadService* if the network conditions are met for auto-upload.

Issue #262, opened 2017-08-19
Audio starts when stopping and alarm or a video in chrome.


megabitdragon:
Steps to reproduce:

1. Start a song in the Amahi app and pause it.
2. Open a browser and watch a movie (e.g.: facebook)
3. Once the movie ends, the song in amahi starts playing

PR #263, opened 2017-08-19
Fixed auto start of song on audio focus gain

octacode:


src/main/java/org/amahi/anywhere/service/AudioService.java
```diff
@@ -409,10 +409,6 @@ public void onAudioFocusChange(int audioFocus) {
 	private void handleAudioFocusGain() {
 		if (isAudioPlaying()) {
 			setUpAudioVolume();
-		} else {
-			if (audioFocus == AudioFocus.LOSS) {
-				playAudio();
```
csoni111: Removing this will break the functionality. Suppose a song was playing and you receive a call, now due to focus loss the song will be paused. But as soon as the call ends, the song must resume playing. This will not happen if we remove this line.

csoni111:
Resolves #262

Issue #264, opened 2017-08-19
App crashes while playing music and pressing the home button


octacode:
#### Steps to reproduce:-
Start a music with album art and then press the home button. Happens with local HDA the remote server music files works fine.

#### Log for the crash
```FAILED BINDER TRANSACTION !!!  (parcel size = 1441816)
08-19 21:19:17.659 18233-18233/org.amahi.anywhere E/AndroidRuntime: FATAL EXCEPTION: main
                                                                    Process: org.amahi.anywhere, PID: 18233
                                                                    java.lang.RuntimeException: android.os.TransactionTooLargeException: data parcel size 1441816 bytes
                                                                        at android.app.ActivityThread$StopInfo.run(ActivityThread.java:3776)
                                                                        at android.os.Handler.handleCallback(Handler.java:751)
                                                                        at android.os.Handler.dispatchMessage(Handler.java:95)
                                                                        at android.os.Looper.loop(Looper.java:154)
                                                                        at android.app.ActivityThread.main(ActivityThread.java:6123)
                                                                        at java.lang.reflect.Method.invoke(Native Method)
                                                                        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:867)
                                                                        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:757)
                                                                     Caused by: android.os.TransactionTooLargeException: data parcel size 1441816 bytes
                                                                        at android.os.BinderProxy.transactNative(Native Method)
                                                                        at android.os.BinderProxy.transact(Binder.java:615)
                                                                        at android.app.ActivityManagerProxy.activityStopped(ActivityManagerNative.java:3700)
                                                                        at android.app.ActivityThread$StopInfo.run(ActivityThread.java:3768)
                                                                        at android.os.Handler.handleCallback(Handler.java:751) 
                                                                        at android.os.Handler.dispatchMessage(Handler.java:95) 
                                                                        at android.os.Looper.loop(Looper.java:154) 
                                                                        at android.app.ActivityThread.main(ActivityThread.java:6123) 
                                                                        at java.lang.reflect.Method.invoke(Native Method) 
                                                                        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:867) 
                                                                        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:757)```

octacode:
Fixed by #265

PR #265, opened 2017-08-19
Resolve app crash due to TransactionTooLargeException on android 7

csoni111:
Removed saving and retrieving data from instance state and shifted it to load from audio service instead.


csoni111:
resolves #264

PR #272, opened 2017-08-28
New apis for upload & delete and auto upload feature

csoni111:


cpg:
The previous pull may have created a conflict

PR #277, opened 2017-10-15
Added code to delete file if it already exists.

rizal-rovins:
The file is being downloaded to /storage/emulated/0/Download/  
This code checks if a file of same name is already present in this location.
If it is present ,then the file is deleted and downloading starts.


src/main/java/org/amahi/anywhere/util/Downloader.java
```diff
@@ -66,6 +67,12 @@ private void setUpDownloadReceiver() {
     }
 
     private void startDownloading(Uri downloadUri, String downloadName) {
+        
+        ///code to delete the file if it already exists
+        File file = new File(Environment.getExternalStorageDirectory()+"/"+Environment.DIRECTORY_DOWNLOADS+"/"+downloadName);
+        if(file.exists());
```
csoni111: This is an empty if statement. That is it does not make any difference if the condition is true or false, the line below this will be called whatsoever.
Note the `;` you have put at the end of if statement.

src/main/java/org/amahi/anywhere/util/Downloader.java
```diff
@@ -66,6 +67,12 @@ private void setUpDownloadReceiver() {
     }
 
     private void startDownloading(Uri downloadUri, String downloadName) {
+        
+        ///code to delete the file if it already exists
+        File file = new File(Environment.getExternalStorageDirectory()+"/"+Environment.DIRECTORY_DOWNLOADS+"/"+downloadName);
```
csoni111: You are checking for the file in the wrong directory. The file is saved in `Android/data/org.amahi.anywhere/files/Download/filename.ext` where as you are looking in `Downloads/filename.ext`. Check the docs here: [Environment.getExternalStorageDirectory()](https://developer.android.com/reference/android/os/Environment.html#getExternalStorageDirectory()) and [Context.getExternalFilesDir(String)](https://developer.android.com/reference/android/content/Context.html#getExternalFilesDir(java.lang.String)).

So replacing `Environment.getExternalStorageDirectory()+"/"+Environment.DIRECTORY_DOWNLOADS` with `context.getExternalFilesDir(Environment.DIRECTORY_DOWNLOADS)` will do the job!

src/main/java/org/amahi/anywhere/util/Downloader.java
csoni111: Also please format the code properly before committing. Check this answer [here](https://stackoverflow.com/a/16580200/6352482).

rizal-rovins:
Deeply sorry for the carelessness on my part.
I have made the requested changes.

cpg:
Should we not do this against `beta` (not `master`)

csoni111:
Yeah right, changing the base branch for the pr.

cpg:
great job!

Issue #255, opened 2017-08-03
Upload stops when the device is rotated


octacode:
Here's video which describes the case: 
https://drive.google.com/open?id=0B3goqm-HUlJvejRZRnV1bVJ2ZmM

sd1998:
@csoni111  are you still working on the issue?

csoni111:
@sd1998 you may work on it if you would like!

sd1998:
@csoni111  could you please point out the file where this error is occurring?

csoni111:
The problem is here in [ServerFilesActivity.java:447](https://github.com/amahi/android/blob/7577b17482552689e4c75dc348db8e2faf67985b/src/main/java/org/amahi/anywhere/activity/ServerFilesActivity.java#L447). Here we are showing a progress dialog but on screen rotation the activity is recreated and so the dialog is dismissed. Changing it to a DialogFragment would solve the problem. Check this [answer](https://stackoverflow.com/a/15729764/6352482).

sd1998:
@csoni111 Thanks for the help :)

sd1998:
I will send a pull request by tomorrow

cpg:
@sd1998 please write support at amahi dot org asking for the developer API. In IRC people come and go.

ArihantJain456:
I would like to work on this issue...

csoni111:
@ArihantJain456 sure, you may work on this. Also you may take reference from the previous pr submitted #282 .

PR #282, opened 2017-10-24
Fix for upload stopping when the device is rotated (#255)

sd1998:
Used DialogFragment to fix Issue #255

### DIFF
--- a/src/main/java/org/amahi/anywhere/activity/ServerFilesActivity.java
+++ b/src/main/java/org/amahi/anywhere/activity/ServerFilesActivity.java
@@ -442,7 +442,7 @@ public class ServerFilesActivity extends AppCompatActivity implements EasyPermis
 
     private void uploadFile(File uploadFile) {
         serverClient.uploadFile(0, uploadFile, getShare(), file);
-        new ProgressDialogFragment();
+        new ProgressDialogFragment().show(getFragmentManager(),"tag");
     }
 
     @Subscribe


src/main/java/org/amahi/anywhere/activity/ServerFilesActivity.java
```diff
@@ -64,6 +65,7 @@
 import org.amahi.anywhere.util.Fragments;
 import org.amahi.anywhere.util.Intents;
 import org.amahi.anywhere.util.Mimes;
+import org.videolan.libvlc.Dialog;
```
csoni111: Also please remove unnecessary imports and reformat the code.

src/main/java/org/amahi/anywhere/activity/ServerFilesActivity.java
```diff
@@ -555,4 +553,22 @@ private void tearDownFilesState(Bundle state) {
         private State() {
         }
     }
+
+    @SuppressLint("ValidFragment")
```
csoni111: Why did you suppress the lint? It's giving a fair warning:
image

Also the app **crashes** on screen rotation with this in the logs:
` java.lang.RuntimeException: Unable to start activity ComponentInfo{org.amahi.anywhere/org.amahi.anywhere.activity.ServerFilesActivity}: android.app.Fragment$InstantiationException: Unable to instantiate fragment org.amahi.anywhere.activity.ServerFilesActivity$ProgressDialogFragment: make sure class name exists, is public, and has an empty constructor that is public.`

cpg:
@sd1998 any updates?

sd1998:
@cpg I could not concentrate on this due to work load in college I will be working on it now

PR #288, opened 2018-01-12
Metadata cache

csoni111:
**Following things have been done:**

- Memory and Disk cache for caching audio metadata for faster loading of song metadata like album art, title etc.
- [Chuck Interceptor](https://github.com/jgilfelt/chuck) for easy debugging of all the api calls

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -39,9 +39,9 @@ android {
 
 
     defaultConfig {
-        def versionMajor = 2
-        def versionMinor = 9
-        def versionPatch = 4
+        def versionMajor = 3
+        def versionMinor = 0
+        def versionPatch = 0
 
         versionName buildVersionName(versionMajor, versionMinor, versionPatch)
         versionCode buildVersionCode(versionMajor, versionMinor, versionPatch)
@@ -145,6 +145,8 @@ dependencies {
     androidTestCompile 'com.android.support:support-annotations:25.4.0'
     provided 'com.squareup.dagger:dagger-compiler:1.2.5'
     compile 'com.github.wseemann:FFmpegMediaMetadataRetriever:1.0.14'
+    debugCompile 'com.readystatesoftware.chuck:library:1.1.0'
+    releaseCompile 'com.readystatesoftware.chuck:library-no-op:1.1.0'
 }
 
 task generateWrapper(type: Wrapper) {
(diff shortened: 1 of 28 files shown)


cpg:
How is the interceptor being invoked?

Will everybody see it? (I don't see code that checks if it's some sort of debug mode)

cpg:
@csoni111 How is the interceptor being invoked?

Will everybody see it? (I don't see code that checks if it's some sort of debug mode)

build.gradle
```diff
@@ -145,6 +145,9 @@ dependencies {
     androidTestCompile 'com.android.support:support-annotations:25.4.0'
     provided 'com.squareup.dagger:dagger-compiler:1.2.5'
     compile 'com.github.wseemann:FFmpegMediaMetadataRetriever:1.0.14'
+    debugCompile 'com.readystatesoftware.chuck:library:1.1.0'
+    releaseCompile 'com.readystatesoftware.chuck:library-no-op:1.1.0'
+    betaReleaseCompile 'com.readystatesoftware.chuck:library-no-op:1.1.0'
```
csoni111: In here for debug app, the dependency is `library` whereas for release and betaRelease variants the dependency is `library-no-op`. The no-op variant of the library contains no actual code just the empty function definitions. So that the code does not crash and also that unused code is not compiled into the apk. So whenever we build the debug app everyone will see it, but in release app it will have no effect.

src/main/java/org/amahi/anywhere/server/ApiModule.java
```diff
@@ -46,10 +47,11 @@
 public class ApiModule {
     @Provides
     @Singleton
-    OkHttpClient provideHttpClient(ApiHeaders headers, HttpLoggingInterceptor logging) {
+    OkHttpClient provideHttpClient(ApiHeaders headers, HttpLoggingInterceptor logging, ChuckInterceptor chuck) {
         OkHttpClient.Builder clientBuilder = new OkHttpClient.Builder();
         clientBuilder.addInterceptor(headers);
         clientBuilder.addInterceptor(logging);
+        clientBuilder.addInterceptor(chuck);
```
csoni111: We can add a check here to add the interceptor only when we have to debug the api.

PR #293, opened 2018-02-28
Fix for handling back press on Authentication Activity (#292)

### LINKED ISSUE #292: Getting Introduction Activity on pressing back button
Introduction Activity is not being shown on the first launch when the user is not logged in. It gets added to Back Stack and is replaced by Authentication Activity.

Can anyone confirm that this is a bug or done intentionally?

adi23arora:
This fixes #292

### DIFF
--- a/src/main/java/org/amahi/anywhere/activity/AuthenticationActivity.java
+++ b/src/main/java/org/amahi/anywhere/activity/AuthenticationActivity.java
@@ -308,9 +308,9 @@ public class AuthenticationActivity extends AccountAuthenticatorActivity impleme
 
         BusProvider.getBus().unregister(this);
     }
-    
+
     @Override
-    public void onBackPressed(){
+    public void onBackPressed() {
         finishAffinity();
     }
 }
--- a/src/main/java/org/amahi/anywhere/activity/AuthenticationActivity.java
+++ b/src/main/java/org/amahi/anywhere/activity/AuthenticationActivity.java
@@ -308,4 +308,9 @@ public class AuthenticationActivity extends AccountAuthenticatorActivity impleme
 
         BusProvider.getBus().unregister(this);
     }
+    
+    @Override
+    public void onBackPressed(){
+        finishAffinity();
+    }
 }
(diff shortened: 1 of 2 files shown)


src/main/java/org/amahi/anywhere/activity/AuthenticationActivity.java
```diff
@@ -308,4 +308,9 @@ protected void onPause() {
 
         BusProvider.getBus().unregister(this);
     }
+
+    @Override
+    public void onBackPressed() {
+        finishAffinity();
```
csoni111: I believe this is unnecessary.

src/main/java/org/amahi/anywhere/activity/IntroductionActivity.java
```diff
@@ -59,5 +59,6 @@ public void onSkipPressed(Fragment currentFragment) {
 
     private void launchTv() {
         startActivity(new Intent(this, NavigationActivity.class));
```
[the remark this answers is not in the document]
adi23arora: I also wondered why it was done this way?

src/main/java/org/amahi/anywhere/activity/AuthenticationActivity.java
adi23arora: There are 2 cases : 
1. When the user is not logged in and its also the first time launch - In this case, we will have Introduction Activity and Navigation Activity in the back stack. On pressing back button first Introduction Activity will be popped then on second press Navigation Activity (with a blank screen).
2. When the user is not logged in and not a first time launch - In this, we would have Navigation Activity in back stack.

So this finishes all activities in stack and works well both of these cases

adi23arora:
@csoni111 Combination of these 2 commits worked well to solve all the problems arising. If there's any other way or approach to this, I would love to discuss and implement.

src/main/java/org/amahi/anywhere/activity/AuthenticationActivity.java
csoni111: Ok I understand the use cases. This seems fine.

Issue #296, opened 2018-03-01
Items on share screen doesn't show any pressing effect


mohitsvnit:
Items on below screen like "Demo" doesn't show any pressing effect on pressing.

![screenshot_2018-03-01-19-23-12](https://user-images.githubusercontent.com/12980721/36848225-a4c825b2-1d86-11e8-834a-fa40294fe993.png)

csoni111:
I tried, it seems to show the ripple effect on press.
![demo screen](https://i.imgur.com/IPDKK6t.jpg)

mohitsvnit:
I am building beta branch app.
My OS: Android 5.1

![screenshot_2018-03-01-19-37-34](https://user-images.githubusercontent.com/12980721/36848824-64f80c20-1d88-11e8-968f-a07741ddea4c.png)

adi23arora:
@mohitsvnit, I too can see the ripple effect. I guess this ripple effect works only on API 21 onwards.

mohitsvnit:
My os version is 5.1 so its api level is 22

mohitsvnit:
I solved the this issue. Should I commit change?

![screenshot_2018-03-01-21-55-00](https://user-images.githubusercontent.com/12980721/36856215-7605166c-1d9b-11e8-85cc-3518bd2f7e0e.png)

csoni111:
Closed via #305

PR #305, opened 2018-03-03
 Pressing effect on items of share fragment. Issue #296

mohitsvnit:
Currently the code used for implementing pressing effect is as below

`<TextView xmlns:android="http://schemas.android.com/apk/res/android"
    android:id="@+id/text"
    android:layout_height="?android:attr/listPreferredItemHeight"
    android:background="?android:attr/activatedBackgroundIndicator"
    android:foreground="?android:attr/selectableItemBackgroundBorderless"
    android:gravity="center_vertical"
    android:maxLines="1"
    android:padding="8dp"
    android:textSize="@dimen/text_medium" />`

In the above code, the property of "foreground" is only supported from api level 23. Hence all the device with api level below 23 doesn't show any pressing effect.

I modified the code and now pressing effect can be seen also in device below 23 as well as above 23.
I tested them in android os 5, 6, 7, 8. 

- os 5
android5

- os 6
android6

- os 8
android8


csoni111:
As @mohitsvnit suggested, Foregrounds were added to **View** class as of Android 6.0 (API 23) but they are available in **ViewGroup** class (specifically FrameLayout) in API 21 too. So instead of going into all the trouble of passing context and setting background programatically, I have another shorter way to resolve this.
If we wrap the TextView inside a FrameLayout and set foreground property to it like this:
```
    android:foreground="?android:attr/selectableItemBackgroundBorderless"
```
then it works just fine on both below and above 23.

csoni111:
Upon more testing I figured out two more solutions:
**1) Replacing these two lines in TextView:**
```
    android:background="?android:attr/activatedBackgroundIndicator"
    android:foreground="?android:attr/selectableItemBackgroundBorderless"
```
with these:
```
    android:background="?android:attr/selectableItemBackground"
    android:foreground="?android:attr/selectableItemBackgroundBorderless"
```
does the job. The background attribute takes care of below 23 and foreground one for above 23.

**2) We could simply make two different layout files:**
`layout\view_server_share_item.xml` with this line:
```
    android:background="?android:attr/selectableItemBackground"
```
`layout-v23\view_server_share_item.xml` with this line:
```
    android:foreground="?android:attr/selectableItemBackgroundBorderless"
```

csoni111:
I think the best way would be 3rd one. @mohitsvnit What do you suggest?

mohitsvnit:
@csoni111  I also tested method 2 and method 3. But in both these method you won't observer ripple effect(on pressing) below api 23. one will observe normal pressing effect. I didn't know about frame layout approach. Did you tested it? If not then I would love to try that.

mohitsvnit:
As you suggested @csoni111. I created used method 3 and created another xml file in v23 layout folder.

I did one change in that. Instead of using `?android:attr/selectableItemBackground`, I used `@drawable/background_ripple_view_server_share_item` so that we get ripple effect also in below api 23.

I tested the code in android os 5,6,7,8. It's working in all of them.

csoni111:
@mohitsvnit I re-tested the 3rd method and it seems to work fine on both a real device and an emulator (running android 5.1). It shows the exact same ripple effect as shown on API 23 and above. I have pushed the code changes on my fork. Please test the code from [here](https://github.com/csoni111/amahi-android/tree/ripple-v23) and let me know.

mohitsvnit:
Yes it's working.

mohitsvnit:
I will revert the commits.

mohitsvnit:
@csoni111 Should I change my code and commit again or cancel this PR?

mohitsvnit:
I still kept textview enclosed in linearlayout in case in future one need to add more views inside it. Is it ok or I remove it?

csoni111:
You must remove the LinearLayout. It is creating the extra load of inflating one more view per list item. we can add it later on if and when required.

mohitsvnit:
Done.

Issue #311, opened 2018-03-06
Displaying last opened time of each file


octacode:
We'd like to display the last time the file was opened of the file to the user. The most feasible approach for this is to show the last opened time in the ListView. 

Also for this I'd recommend using SQLite with three entries:
1. Path (Stores the path of the file).
2. Time.
3. Date.

adi23arora:
@octacode So, how are we planning to implement this, using Android's SQLite APIs or Room Persistence Library?
I would prefer the latter because it involves less boilerplate code and is less time-consuming.

octacode:
Room Persistence Library is good option, however the present code uses Android's SQLite APIs and also including Room will increase the apk size, and we don't really want it cause we already are at 80.1 MB :open_mouth:

adi23arora:
Yes, the apk size might be a concern here.
So, where exactly can the last open date-time is be placed in ListView?

octacode:
In the list we are already displaying the last modified time which is fetched from the server, Honestly I want to keep that. What we can do here is that `onLongPress` we can show an info icon in the activated action bar. Clicking on which will open an alert dialog box with title as Info and it will contain the following details:
1. Name of the file
2. Mime Type
3. Size of the file
4. Last opened
5. Last modified

@cpg @csoni111 How does this sound to you? Also @adi23arora should I assign this to you?

adi23arora:
@octacode Yes, please assign me, I want to work on this.

octacode:
Assigned to @adi23arora .

Somehow I wasn't able to see his name in the Assignees, So for the time being this will do.

@cpg can you please look into the reason for this^.

csoni111:
How about instead of storing the path of the file, we store a unique key for each file/folder. This key can be generated using some fields from its metadata like name, last_modified etc. [This](https://github.com/amahi/android/blob/a963e50d78ba3ce5f99fd00ae6fb4bc0923f75d6/src/main/java/org/amahi/anywhere/server/model/ServerFile.java#L114:L129) can be reused if it serves the purpose.

Also instead of storing the date and time separately, it will be easier to store just the timestamp/epoch in Long.

octacode:
1. Yes we can store the timestamp instead of date and time separately :+1: 
2. Why complicate it for us? Instead of having some keys which definitely are not going to be readable since they are comp generated we can have the path in the db entries (will make debugging a bit easier for us). Also there'll be overload of creating keys while we have direct access to the path of the file. 

This way we'll have two entries:
[...]

PR #313, opened 2018-03-10
Add visual cues for played/unplayed (#311)

adi23arora:
### Agreement with DB requirements:

This describes all the requirements the DB interface need to have, as quoted under https://github.com/amahi/android/issues/315.

1. **Resilient to corruption** - I have re-tested the functions and the operations are done in a serialized manner which avoids any kind of conflicts in the database. Database operations are carried out using the `FileInfoDbHelper.java` class which prevents direct access to `FileInfoDb.java` class from outside the package.
Any operation that could have led to an exception is handled using a try-catch block and null checks are also there to prevent any null pointer exceptions. So, this ensures that database can't be corrupted from any internal bugs.

2. **Self Stabilizing** - If in case any failure occurs leading to database corruption then it is very swiftly handled by the `DefaultDatabaseErrorHandler` class present with the Android's database API which handles any SQLite database corruption by default. So, this makes it self-stabilizing without any external help or intervention.
Here's a reference to the [docs](https://developer.android.com/reference/android/database/DefaultDatabaseErrorHandler.html).

3. **Versatile** - Currently, we are using it for the visual cues for played and unplayed media files. But it can be further extended or upgraded to add more columns, tables to support multiple other features and store some other useful data regarding files. 

4. **Upgradable** - `FileInfoDb.java` extends `SQLiteOpenHelper` class which has an abstract method `onUpgrade` that can be overridden to handle database upgrades, further we have a `DATABASE_VERSION` variable which is currently set (=1) but it can be incremented with every new upgrade to database, to support database migration and modification.
It can be modified to handle the upgrade to different database versions and currently it is set to follow the general approach for upgrades.
(Downgrades are not required as such but we have a similar function `onDowngrade` to handle downgrades which are currently left blank.)

5. **Transparent** - Pre-defined database operations are capable of handling around 1000s' of operations per second. But here we are performing only a single operation for each file accessed which takes place in a very small fraction of a second, so there won't be any delay or lag with such operations.
And even if we want to perform those 1000s of batch operations, those can also be optimized using transactions.

6. **Consistent** - All the data like the file paths are collected from the backend APIs helper methods and fed into the database, so it is consistent with the current information we are receiving from the database.

Moreover, this database is implemented with a different name so it won't interfere with any previous databases used in our app. And it is just responsible for some new additional features and it won't be affecting the overall purpose or the sole functioning of `Amahi` app in any way.

### DIFF
--- a/src/main/java/org/amahi/anywhere/adapter/ServerFilesAdapter.java
+++ b/src/main/java/org/amahi/anywhere/adapter/ServerFilesAdapter.java
@@ -46,6 +46,7 @@ import org.amahi.anywhere.server.model.ServerShare;
 import org.amahi.anywhere.task.AudioMetadataRetrievingTask;
 import org.amahi.anywhere.util.Mimes;
 import org.amahi.anywhere.util.RecyclerViewItemClickListener;
+import org.amahi.anywhere.util.Preferences;
 
 import java.text.SimpleDateFormat;
 import java.util.Collections;
@@ -134,7 +135,7 @@ public class ServerFilesAdapter extends FilesFilterAdapter {
 
             setUpDbHelper();
 
-            String file_path = serverShare.getName() + file.getPath();
+            String file_path = Preferences.getServerName(context) + "/" + serverShare.getName() + file.getPath();
 
             if (fileInfoDbHelper.getFilePlayed(file_path)) {
                 fileHolder.fileTextView.setTypeface(fileHolder.fileTextView.getTypeface(), Typeface.BOLD);
--- a/src/main/java/org/amahi/anywhere/adapter/ServerFilesAdapter.java
+++ b/src/main/java/org/amahi/anywhere/adapter/ServerFilesAdapter.java
@@ -22,6 +22,7 @@ package org.amahi.anywhere.adapter;
 import android.content.Context;
 import android.graphics.Bitmap;
 import android.support.v7.widget.RecyclerView;
+import android.graphics.Typeface;
 import android.text.Spannable;
 import android.text.SpannableStringBuilder;
 import android.text.TextUtils;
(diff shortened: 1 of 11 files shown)


cpg:
Thanks for this. We're not ready for this, I think.

Any time there is state, it has to have certain properties, in order to not lose sanity. I wanted to specify them separately as a "DB requirements" issue, but this is coming in sooner than I expected. So here is a quick stab a DB interface requirements:

  * **resilient to corruption**: it _must not_ (repeat _must not_) ever be corrupted. hardware errors or even bugs could cause this, but it definitely should be very very very rare an not self-inflicted by our code
  * **self-stabilizing**: if it ever does get corruption, it either has to automatically be detected an fixed quickly and swiftly, or it has to be re-initialized quickly so has it works well.
 * **versatile**: has to be used for multiple purposes
 * **upgradeable**: it has to be able to be upgradeable automatically, migrating from one schema to the next and with proper data migration. The right hooks have to be in the app to upgrade (and downgrade?) if new uses are needed or formats changed 
 * **transparent**: it must not affect app performance.

Should this be tackled separately?

octacode:
The PR was a bit quick. Great job there @adi23arora. 

I did a quick test of this and it looks fine to me. However I'd like some more time to test it thoroughly and there are some cases where we should decide how the database should behave, some of which are:-
#### 1. When user logs out and logs in with a new id:
In this case I suggest we should delete the previous table. 
#### 2. Let's say I have set up two servers and in both of them we have the same file at the same path then it would be a bit trouble some, for this we should have the server name appended in the path:
from `/server_file/alpha.mp4` to `server_name/server_folder/server_file`.

There might be some more unseen issues which need to be addressed.

adi23arora:
@cpg It won't affect any of the present features, but still I would re-test to make sure it is in accordance with the quoted DB requirements.

@octacode Thanks for the review, I would make the changes as requested and would try to look for more such use cases that need to be addressed.

cpg:
@adi23arora please note that `would` != `will`. Did you mean to use would in your last comment?

I still would like a point-by-point reasoning in depth that this implementation will satisfy every single one of the requirements for the DB. I created a separate issue for the DB stuff.

I don't think we should merge this in until we have a thorough review vs the requirements.

adi23arora:
@cpg I have modified my previous comment. I hope this time I am clear and you understand my willingness to work on this. I **will** get back with the changes and proper reasoning to support my work.

adi23arora:
@cpg I have made the changes as requested and added an explanation (in description) to explain all the DB requirements (https://github.com/amahi/android/issues/315), as promised. Please have a look at it.

@octacode @csoni111 Please review the changes and give feedback on the same.

cpg:
I would like to discuss in a mentors hangout because I think this area is particularly risky. However, it make take a little while.

This should probably be an item in your GSoC proposal @adi23arora.

cpg:
Could we resolve the conflicts and push this one to the `database` branch?

cpg:
Merged into `database` branch. Great job @adi23arora and great point-by-point explanation.

Issue #307, opened 2018-03-04
app throws error while accessing shares of own server on android phone


mohitsvnit:
When I access shares of "Welcome to amahi" server it works fine.
But when I access shares of my own server, It shows connection error.

octacode:
Can you put the logs of the error here?

mohitsvnit:
java.net.ProtocolException: Too many follow-up request: 21

octacode:
Was not able to replicate the issue. Closing this for now

mohitsvnit:
I can upload video of it if you like.

csoni111:
@mohitsvnit you must be missing the installation of Amahi Anywhere app on your hda. Got to apps tab on hda dashboard and install `Amahi Anywhere` from the list to resolve this error.

cpg:
We should handle this error better in the app, so as the user is not left confused.

While it may not be common that the AA server is not running, it can happen. The app should do the best of it, I think.

It should be easy to test by stopping the server for a bit and handle it with a nice, clear, message in the app.

mohitsvnit:
@csoni111 My amahi server was running. But you are right I didn't installed any app in it yet.
@cpg Is there any way to check if `Amahi Anywhere app` is running or not?

mohitsvnit:
@csoni111 You were right. It's an problem of not running `Amahi anywhere`.
Also I found out that in `ServerShareFragment` there is not action if  `serverClient` is not connected while getting shares. That is what happening if `Amahi anywhere app` is not running. So progress dialog keep rotating for infinite time. 
@cpg Should I add error msg in errorLinearLayout or show that in `SnackBar`?

mohitsvnit:
`serverClient` was not connecting because `ServerRouteResponse` is failing. There is no action when that is happening.
I am thinking to add toast with message `Your Amahi server may not be running Amahi Anywhere. Please ensure the Amahi Anywhere app is installed and running.` when `ServerRouteResponse` is failed. 
@cpg @csoni111  what do you think?

PR #319, opened 2018-03-12
Fixed error of infinite loading when `amahi anywhere` app is not running at hda server and added feature to reconnect to server by click on retry. Issue #307

mohitsvnit:
Fixed error of infinite loading when `amahi anywhere` app is not running at hda server.
Also now user can initiate reconnect to server by clicking retry button on error screen.

I tried to follow amahi code structure for creating events and flow of events.

### DIFF
--- a/src/main/java/org/amahi/anywhere/bus/ServerConnectionFailedEvent.java
+++ b/src/main/java/org/amahi/anywhere/bus/ServerConnectionFailedEvent.java
@@ -20,4 +20,15 @@
 package org.amahi.anywhere.bus;
 
 public class ServerConnectionFailedEvent implements BusEvent {
+    private String errorMessage;
+
+    public ServerConnectionFailedEvent(){}
+
+    public ServerConnectionFailedEvent(String errorMessage) {
+        this.errorMessage = errorMessage;
+    }
+
+    public String getErrorMessage() {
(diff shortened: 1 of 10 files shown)


mohitsvnit:
@cpg @csoni111 If code convention doesn't follow amahi code convention then please suggest changes.

mohitsvnit:
@octacode I found that if `Amahi Anywhere` app is not running on server then request to get server route will fail. Is there any other reason possible for that? 
I was facing this problem from many days. If @csoni111  didn't suggested said that maybe `Amahi Anywhere` app is not running. Maybe somewhere it needs to mention. 
If you are saying I will remove toast.

octacode:
@cpg your views on whether we should keep the meesage?

cpg:
Are you thinking that the app should retry indefinitely without user feedback? ... I'm not too sure about that.

In the bug, I mentioned that it may be rare, but if there is a bug or a networking issue, the user may just not be aware that this could be the issue. Or at least imply the AA server is not reachable, which may indicate a networking issue, not necessarily a networking issue. I changed to message to include reachability as a possible issue
```
The Amahi Anywhere app in your server is not reachable. Please ensure that Amahi Anywhere app is installed, running and reachable in your network.
```

PR #320, opened 2018-03-12
Uri permissions for open files intent

### LINKED ISSUE #290: trouble opening PDF files
user reporting in ..

> Using Samsung Galaxy S7 Android phone.
> Amahi for Android app displays all files and plays mp3s but will not open text files such as PDF. I do have Adobe on my phone. The app reports it downloads the file but then reports it cannot be accessed, check network or location.
> This happens whether connecting via LAN or remotely.

ArihantJain456:
Now files like pdf, doc, txt, xls, etc can be opened directly with intent.

Closes #290


ArihantJain456:
@cpg give your suggestions regarding changes if any.

PR #328, opened 2018-03-13
added logic for correct working of auto upload (#321)

### LINKED ISSUE #321: Auto upload stuck disabled
I am not sure how I got into this situation in the app, but I tried enabling it, which worked, then I think I maybe had WiFi disabled and the app was having trouble accessing my server list. I had to close the app, the enabled WiFi and ran the app again.

Servers were there, browsed about and tested a bugfix. Went to Settings > AutoUpload and cannot enable it any more. Restarting the app does not not help, seems permanently stuck.

This means there is some state somewhere and it's not managed well!

ArihantJain456:
This PR implements two things:
1. Proper working of  Auto Upload Toggle Button in `UploadSettingsFragment`.
2. Allows Auto Upload option to update when coming back in `SettingsFragment` from `UploadSettingsFragment`.

Changes After Updating PR:
1. Added summary for auto uploading settings.
2. Changed title from `Auto Upload` to `Auto Upload Photos`
3. Enabled working in all three mode namely `LAN`, `Auto Detection` and `Remote`.
closes #321

### DIFF
--- a/src/main/java/org/amahi/anywhere/fragment/UploadSettingsFragment.java
+++ b/src/main/java/org/amahi/anywhere/fragment/UploadSettingsFragment.java
@@ -39,7 +39,6 @@ import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 import android.support.annotation.RequiresApi;
 import android.support.design.widget.Snackbar;
-import android.util.Log;
 import android.view.View;
 
 import com.squareup.otto.Subscribe;
@@ -83,6 +82,7 @@ public class UploadSettingsFragment extends PreferenceFragment implements
     ServerClient serverClient;
 
     private String authenticationToken;
+    private ArrayList<Server> activeServers;
 
     @Override
     public void onCreate(@Nullable Bundle savedInstanceState) {
@@ -106,33 +106,27 @@ public class UploadSettingsFragment extends PreferenceFragment implements
     private void setUpSettings() {
         setUpSettingsContent();
         setUpSettingsTitle();
-        toggleUploadSettings(isUploadEnabled());
+        setUpAuthenticationToken();
         setUpSettingsListeners();
     }
 
     private void setUpSettingsContent() {
         addPreferencesFromResource(R.xml.upload_settings);
-        setUploadPathSummary(getUploadPath());
+        setUploadLocationSummary(getUploadLocation());
     }
 
-    private void setUploadPathSummary(String path){
-        if(!path.isEmpty()){
-            if(path.length()>25){
-                path = path.substring(0,21);
-                path += "...";
+    private void setUploadLocationSummary(String location) {
+        if (!location.isEmpty()) {
+            if (location.length() > 25) {
+                location = location.substring(0, 21);
+                location += "...";
             }
-            getPathPreference().setSummary(path);
+            getLocationPreference().setSummary(location);
         }
     }
 
     private void setUpSettingsTitle() {
         getAutoUploadSwitchPreference().setTitle(getAutoUploadTitle(isUploadEnabled()));
-
-        if (isConnectionLocal()) {
-            getHdaPreference().setTitle(getString(R.string.preference_title_upload_server));
-        } else {
-            getHdaPreference().setTitle(getString(R.string.preference_title_upload_hda));
-        }
     }
 
     private AccountManager getAccountManager() {
@@ -175,12 +169,12 @@ public class UploadSettingsFragment extends PreferenceFragment implements
 
     @Subscribe
     public void onServersLoaded(ServersLoadedEvent event) {
-        if(isUploadEnabled())
         setUpServersContent(event.getServers());
     }
 
     private void setUpServersContent(List<Server> servers) {
-        ArrayList<Server> activeServers = filterActiveServers(servers);
+
+        activeServers = filterActiveServers(servers);
         String[] serverNames = new String[activeServers.size()];
         String[] serverSessions = new String[activeServers.size()];
 
@@ -196,9 +190,12 @@ public class UploadSettingsFragment extends PreferenceFragment implements
 
         String session = getHdaPreference().getValue();
         if (session != null) {
-            setUpServer(session);
-            String selectedServerName = getHdaPreference().getEntry().toString();
-            getHdaPreference().setSummary(selectedServerName);
+            Server server = getServer(session);
+            if (server != null) {
+                setUpServer(server);
+                String selectedServerName = getHdaPreference().getEntry().toString();
+                getHdaPreference().setSummary(selectedServerName);
+            }
         }
     }
 
@@ -212,6 +209,17 @@ public class UploadSettingsFragment extends PreferenceFragment implements
         }
 
         return activeServers;
+    }
+
+    private Server getServer(String session) {
+
+        // for working with debug server
+        for (Server server : activeServers) {
+            if (server.getSession().equals(session)) {
+                return server;
+            }
+        }
+        return null;
     }
 
     private boolean isUploadEnabled() {
@@ -228,7 +236,7 @@ public class UploadSettingsFragment extends PreferenceFragment implements
         getAutoUploadSwitchPreference().setOnPreferenceChangeListener(this);
         getHdaPreference().setOnPreferenceChangeListener(this);
         getSharePreference().setOnPreferenceChangeListener(this);
-        getPathPreference().setOnPreferenceChangeListener(this);
+        getLocationPreference().setOnPreferenceChangeListener(this);
     }
 
     @Override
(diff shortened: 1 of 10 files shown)


cpg:
I tested this. I think it's not quite working still. I'm on Android 6.0.1, in case that matters.

1) At some point the first label said `Disable` but the toggle was off. After playing with it now I see it says `Disable` and the toggle is on. I think this may not be the best label. How about `Auto Upload Photos` ? I am not sure why it was not matching the toggle switch. Maybe a mismatch in initial state? (I tried taking screenshots, but for some reason they are not working on my phone/OS)

2) While I'm in the neighborhood ... the label `HDA Server` should be just `Server`. And `Path` could better be `Location` perhaps?

3) Nice to have: It would be really nice to show the name of the server, share and even path on the right, right-aligned.

Perhaps the location could be limited a bit in length if it goes above, say, 25 chars?  I.e. anything above 22 could be replaced by `...`

I did test the uploads and they worked well.

ArihantJain456:
In settingsfragment autoupload indicates whether it is on or off.
In uploadSettings fragments title of toggle switch is Enable if the autoupload is disable indicating that u can enable it.

ArihantJain456:
` Auto Upload Photos'  title looks good
I  will try to implement 2 and 3 feature as you mentioned above.

ArihantJain456:
@cpg I was checking google material design guidelines and found that all label should be left aligned.

[Material Design guidelines for preferences](https://material.io/guidelines/patterns/settings.html#settings-labels-secondary-text)
@octacode what do you say?

cpg:
I'm not certified in Material Designed-ness (I know, surprising).

So hey, if the setting value gotta under the entry, that sounds good to me. My view was that seeing the current setting might be a good idea. It's not a hard requirement.

ArihantJain456:
Yes it is not a hard requirement. I also noticed  some other issues in UploadSettingsFragment as:
1. Path is empty always.
2. When server connection setting is set to Remote, then  shares does not loaded and hence path.

I will update my PR soon after rectifying all these issues.

ArihantJain456:
@cpg please review changes.

csoni111:
The app **crashes** with this in the logs:

```
FATAL EXCEPTION: main
  Process: org.amahi.anywhere, PID: 3823
  java.lang.RuntimeException: Could not dispatch event: class org.amahi.anywhere.bus.ServersLoadedEvent to handler [EventHandler public void org.amahi.anywhere.fragment.UploadSettingsFragment.onServersLoaded(org.amahi.anywhere.bus.ServersLoadedEvent)]: Attempt to invoke interface method 'java.lang.String java.lang.CharSequence.toString()' on a null object reference
      at com.squareup.otto.Bus.throwRuntimeException(Bus.java:460)
      at com.squareup.otto.Bus.dispatch(Bus.java:387)
      at com.squareup.otto.Bus.dispatchQueuedEvents(Bus.java:368)
      at com.squareup.otto.Bus.post(Bus.java:337)
      at org.amahi.anywhere.server.response.ServersResponse.onResponse(ServersResponse.java:78)
      at retrofit2.ExecutorCallAdapterFactory$ExecutorCallbackCall$1$1.run(ExecutorCallAdapterFactory.java:70)
      at android.os.Handler.handleCallback(Handler.java:789)
      at android.os.Handler.dispatchMessage(Handler.java:98)
      at android.os.Looper.loop(Looper.java:164)
      at android.app.ActivityThread.main(ActivityThread.java:6809)
      at java.lang.reflect.Method.invoke(Native Method)
      at com.android.internal.os.Zygote$MethodAndArgsCaller.run(Zygote.java:240)
      at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:767)
   Caused by: java.lang.NullPointerException: Attempt to invoke interface method 'java.lang.String java.lang.CharSequence.toString()' on a null object reference
      at org.amahi.anywhere.fragment.UploadSettingsFragment.setUpServersContent(UploadSettingsFragment.java:195)
      at org.amahi.anywhere.fragment.UploadSettingsFragment.onServersLoaded(UploadSettingsFragment.java:178)
      at java.lang.reflect.Method.invoke(Native Method)
      at com.squareup.otto.EventHandler.handleEvent(EventHandler.java:89)
```
To replicate the error, just do a clean install -> goto settings -> enable auto upload toggle -> allow permissions -> click on the toggle again -> the app crashes.

src/main/java/org/amahi/anywhere/server/client/ServerClient.java
```diff
@@ -210,6 +210,7 @@ public void connectLocal() {
         }
         this.serverAddress = serverRoute.getLocalAddress();
         this.serverApi = buildServerApi();
+        ServerConnectionDetectingTask.execute(serverRoute);
```
csoni111: Why have you added this line here?
This will automatically choose first local address and then switch to remote address if the ping to local address fails. Adding this dissolves the purpose of having `Local` and `Remote` mode explicitly.

src/main/java/org/amahi/anywhere/server/client/ServerClient.java
```diff
@@ -219,6 +220,7 @@ public void connectRemote() {
         }
         this.serverAddress = serverRoute.getRemoteAddress();
         this.serverApi = buildServerApi();
+        ServerConnectionDetectingTask.execute(serverRoute);
```
csoni111: Same comment as on [line#213](https://github.com/amahi/android/pull/328/files#diff-098fbcd587ec255c747f833af334fe5bR213).

src/main/java/org/amahi/anywhere/fragment/UploadSettingsFragment.java
```diff
@@ -111,10 +112,27 @@ private void setUpSettings() {
...
     }
 
     private void setUpSettingsTitle() {
         getAutoUploadSwitchPreference().setTitle(getAutoUploadTitle(isUploadEnabled()));
+
+        if (isConnectionLocal()) {
+            getHdaPreference().setTitle(getString(R.string.preference_title_upload_server));
+        } else {
+            getHdaPreference().setTitle(getString(R.string.preference_title_upload_hda));
+        }
```
csoni111: I don't think we need two different titles for the same field. `HDA Server` and `Server`.
Make it just `Server` as pointed out in 2nd point of the [comment](https://github.com/amahi/android/pull/328#issuecomment-372926236) by cpg.

src/main/res/values/preferences.xml
```diff
@@ -31,13 +31,18 @@
     <string name="preference_title_account_sign_out">Sign out</string>
     <string name="preference_title_server_connection">Connection</string>
     <string name="preference_title_tell_a_friend">Tell a friend</string>
-    <string name="preference_title_upload_settings">Auto Upload</string>
+    <string name="preference_title_upload_settings">Auto Upload Photos</string>
 
     <string name="preference_title_upload_hda">HDA Server</string>
+    <string name="preference_title_upload_server">Server</string>
     <string name="preference_title_upload_share">Share</string>
     <string name="preference_title_upload_path">Path</string>
```
csoni111: Change this title to `Location` instead of `Path` as pointed out in the comments above.

src/main/res/values/preferences.xml
```diff
@@ -31,13 +31,18 @@
...
     <string name="preference_title_upload_data">Allow on Cellular Data</string>
 
+    <string name="preference_summary_server">Select Server</string>
+    <string name="preference_summary_share">Select Share</string>
+    <string name="preference_summary_path">Select Path</string>
```
csoni111: I don't think these three are used anywhere.

src/main/java/org/amahi/anywhere/fragment/UploadSettingsFragment.java
```diff
@@ -216,17 +238,32 @@ public boolean onPreferenceChange(Preference preference, Object newValue) {
             boolean isUploadEnabled = (boolean) newValue;
             if (isUploadEnabled) {
                 if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
-                    checkReadPermissions();
-                    return false;
+                    boolean b = checkReadPermissions();
+                    Log.d("UploadPer","value = " + b);
```
csoni111: Remove these unnecessary logs.

src/main/java/org/amahi/anywhere/fragment/UploadSettingsFragment.java
ArihantJain456: Sorry, I got confused by the 2nd point of the comment. I thought It means to change the title to `Server` when connecting to local and otherwise `HDA Server`.
I will make it `Server`.

src/main/res/values/preferences.xml
ArihantJain456: ok, thanks. I will change it.

src/main/java/org/amahi/anywhere/fragment/UploadSettingsFragment.java
ArihantJain456: ok, I will remove it.

src/main/res/values/preferences.xml
ArihantJain456: They have been used when no server is selected or when no share is selected for the first time.

ArihantJain456:
#### Issues at present Scenario
1. Auto Upload Feature is enabled even when no server is selected or when no share is selected, this results in failure in uploading photo.
2. ServerClient is Singleton, on changing Server from anywhere (from Navigation or from upload option), will change the Server for all functionalities. So when the server is changed to `Welcome To Amahi` server, it will also change the photos upload server to `Welcome To Amahi Server`.

#### Solution
1. Auto Upload Enable toggle will set to Enable only when Server and Share are selected
otherwise a Snackbar will be displayed indicating to select server and share. For this to look more appropriate the Enable toggle Switch can be present at bottom of setting rather than at the top.
2. Singleton server client issue: this needs discussion and a separate issue should be open for it.

@cpg @csoni111 give your views.

cpg:
Solution 1) seems sensible. Good iea.

Solution 2 is above my pay grade. Maybe @csoni111, @octacode or @megabitdragon can speak more sense about it.

csoni111:
I agree with @cpg, solution 1 seems good. About solution 2, I think it should not be mixed here. Please open another issue for that problem and we will discuss it there.

ArihantJain456:
@csoni111 I have removed those extra lines of code. Suggest any more changes if required.

Issue #329, opened 2018-03-14
Action bar title is not updated with audio change


adi23arora:
In `ServerFileAudioActivity` the file name in action bar title is not updated on switching to next or previous audio file.

![ss_amahi](https://user-images.githubusercontent.com/17095968/37434962-6dee4cfc-2807-11e8-879c-f29f1eafff71.jpg)

adi23arora:
I have fixed this, will send a PR soon.

PR #330, opened 2018-03-14
Update Action Bar title with audio change

adi23arora:
This fixes https://github.com/amahi/android/issues/329.

### DIFF
--- a/src/main/java/org/amahi/anywhere/activity/ServerFileAudioActivity.java
+++ b/src/main/java/org/amahi/anywhere/activity/ServerFileAudioActivity.java
@@ -146,7 +146,11 @@ public class ServerFileAudioActivity extends AppCompatActivity implements
     }
 
     private void setUpAudioTitle() {
-        getSupportActionBar().setTitle(getFile().getName());
+        if (audioService != null) {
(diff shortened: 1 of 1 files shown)


octacode:
has been already fixed by #69

cpg:
I went with #331 by @ArihantJain456 because that change was larger (had additional things in it). I realized it would conflict with this.

Looking at the conflict, there is a possible slight difference in logic (near as I can tell, without knowing the APIs in detail).

Could you and @ArihantJain456 try and see of resolving the issues here and either fix the conflict or close it?

ArihantJain456:
@adi23arora App is working fine and also title is changing on changing audio.
Please check the latest beta branch and verify it. And if you think there is something wrong with the titles then  we will resolve it, otherwise close the PR.

adi23arora:
Ok, I guess @ArihantJain456  have already solved it [here](https://github.com/amahi/android/blob/e832b0eab3abc21095936edd1be69fd78a407b7a/src/main/java/org/amahi/anywhere/activity/ServerFileAudioActivity.java#L199-L205) with a different approach (using a Parcelable array) for storing the audio file meta data used for displaying audio related info.

I will close this PR.

PR #339, opened 2018-03-20
adds Java8 features (#326)

### LINKED ISSUE #326: Java8 features
## Description
We should make use of new Android Studio supported Java8 features like lambda Expressions.
for example:
`button.setOnClickListener((View view) -> 
// do something`

They are easier to use and also helpful in writing cleaner code.

ArihantJain456:
## Description
Anonymous Classes  replaced by lambdas
Method references for calling existing methods
Removes unnecessary code
Updates compileSDKVersion and supports libraries for making use of new API features

closes #326 

@cpg, @octacode please review the PR and suggest changes if any.


octacode:
Looks like the Travis CI is failing. @ArihantJain456 can you please look into it?

mohitsvnit:
It's an error of changing android sdk from 25 to 26.
@ArihantJain456 If you are changing sdk then why don't you change it to latest one?

ArihantJain456:
Yes @mohitsvnit Travis CI failing is due to the same reason. I thought  whether changing API version would be right or not, So I changed it to 26 only because I needed some of its features.

sanjitchak:
@ArihantJain456  Simple error of license agreement of the SDK, i guess

sanjitchak:
@mohitsvnit  Won't it face a compatibility issue if the SDK is upgraded to latest one? Never tried it, so, asking.

mohitsvnit:
@sanjitchak maybe it can create some compatibility issue. I thinks only mentors can give their views if sdk should be updated or not.

ArihantJain456:
@octacode should I update ` .travis.yml` ?

octacode:
Yes we should update it. :+1:

Issue #342, opened 2018-03-22
Option to see file metadata


mohitsvnit:
There is no option to see file metadata.

We can add one icon on app bar to show file meta information when long press on file.

mohitsvnit:
Can I work on this?

adi23arora:
@mohitsvnit All the file related info we are getting from the server is already shown in the file item in the list view itself. So what exactly do you want to display here?
Moreover, it would be useful if we have some file related info to show to the user like "last opened" or "last played". And there's already a discussion going for this on issues #311 and #315.

mohitsvnit:
So you are also working on displaying them on long press or just working on collecting data and storing them?

octacode:
>We can add one icon on app bar to show file meta information when long press on file.

This definitely is not self-explanatory. Please try to explain what you're proposing properly. So that it is understandable.

mohitsvnit:
@octacode When user longpress on file, currently we are getting 2 options on appbar i.e. `Share` and `Delete`. Like that we can add 1 more option(with icon) for `Information`. Which will diplay all the metadata of file. Metadata of file will increase with time.

octacode:
This is similar to what I proposed in this comment https://github.com/amahi/android/issues/311#issuecomment-370800178, I think adding this will do us no harm. However this isn't something which is very necessary or required.

PR #354, opened 2018-03-24
Added feature to view file information in ServerFileFragment

mohitsvnit:
This PR fixes issue #342.

I added an icon to view meta data of file in application tool bar in `ServerFileFragment`.
I used Material Design dialog. We can add more information to display when meta data of file will increase.

screenshot_1521917481

screenshot_1521917471

### DIFF
--- a/src/main/java/org/amahi/anywhere/fragment/ServerFilesFragment.java
+++ b/src/main/java/org/amahi/anywhere/fragment/ServerFilesFragment.java
@@ -354,12 +354,10 @@ public class ServerFilesFragment extends Fragment implements
         dialofFileInfo.setContentView(LayoutInflater.from(getContext())
             .inflate(R.layout.dialog_file_information, null));
         TextView textFileName = (TextView) dialofFileInfo.findViewById(R.id.text_file_name);
-        TextView textFileType = (TextView) dialofFileInfo.findViewById(R.id.text_file_type);
         TextView textFileModification = (TextView) dialofFileInfo.findViewById(R.id.text_file_modification);
         TextView textFileSize = (TextView) dialofFileInfo.findViewById(R.id.text_file_size);
 
         textFileName.setText(checkedFile.getName());
-        textFileType.setText(checkedFile.getMime());
         textFileModification.setText(getLocalDateTime(checkedFile.getModificationTime()));
         textFileSize.setText(Formatter.formatFileSize(getContext(), checkedFile.getSize()));
 
--- a/src/main/java/org/amahi/anywhere/fragment/ServerFilesFragment.java
+++ b/src/main/java/org/amahi/anywhere/fragment/ServerFilesFragment.java
@@ -31,6 +31,7 @@ import android.os.Handler;
 import android.os.Parcelable;
 import android.support.annotation.NonNull;
 import android.support.annotation.RequiresApi;
+import android.support.design.widget.BottomSheetDialog;
 import android.support.design.widget.Snackbar;
 import android.support.v4.app.Fragment;
 import android.support.v4.widget.SwipeRefreshLayout;
@@ -39,6 +40,7 @@ import android.support.v7.widget.GridLayoutManager;
 import android.support.v7.widget.LinearLayoutManager;
 import android.support.v7.widget.RecyclerView;
 import android.support.v7.widget.SearchView;
+import android.text.format.Formatter;
 import android.util.DisplayMetrics;
 import android.view.ActionMode;
 import android.view.LayoutInflater;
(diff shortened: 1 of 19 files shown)


src/main/res/values/strings.xml
```diff
@@ -150,4 +151,11 @@
     <string name="intro_title_6">Ready</string>
     <string name="intro_desc_6">You\'re all set to go. Thanks for using Amahi.</string>
 
+    <!--Dialog File Info-->
+    <string name="dialog_file_name">Name:</string>
+    <string name="dialog_file_type">File Type:</string>
+    <string name="dialog_file_modification">Last Modification:</string>
```
adi23arora: I guess it should be "Last Modified" or just "Modified".

mohitsvnit: oops! My bad.

src/main/res/values/strings.xml
```diff
@@ -42,6 +42,7 @@
     <string name="menu_settings">Settings</string>
     <string name="menu_share">Share</string>
     <string name="menu_delete">Delete</string>
+    <string name="menu_information">Information</string>
```
adi23arora: "Properties" would be an apt word for this menu option as we are displaying file properties.

src/main/java/org/amahi/anywhere/fragment/ServerFilesFragment.java
```diff
@@ -339,6 +349,39 @@ public void onClick(DialogInterface dialog, int which) {
...
+                    actionMode.finish();
+
+                }
+            }
+        });
+
+        dialofFileInfo.show();
+    }
+
+    private String getLocalDateTime(Date modificationTime) {;
```
adi23arora: Spare semicolon `;`

src/main/res/values/strings.xml
mohitsvnit: I checked it with other apps. They are using "Information". Lets mentor give review on this one.

cpg:
Looks like there are two columns. In terms of usability, if the labels on the left column can be right-aligned, it will look better. In addition, it could be improved if the color of the labels on the left column could be light gray and the colon could be removed, improving things.

mohitsvnit:
Sure will do that.

mohitsvnit:
Color: material grey 500
screenshot from 2018-04-02 14-20-55

Color: material grey 600
screenshot from 2018-04-02 14-20-27


@cpg please review above design

cpg:
Also (sorry, I just realized this) ... is the cancel button really necessary? It takes a lot of space.

Wouldn't simply touching the other area be enough the close this area? Or even also touching this area in the message?

Also, if not possible, how about calling it "Close" or just an `X` somewhere or something (so long as it complies with MD).

I can't tell much about the difference other than the edges (the second one seems better in that respect).

mohitsvnit:
If the user will touch on another area then the dialog will close but the file will still be in the selected mode and user can select another file by just touching it.

Whereas if the user clicks on `CANCEL` button then the file will be deselected. 

For the button, I used predefined button theme. 
I will try to make it more user-friendly.

cpg:
What is the significance of the file being selected or deselected? (seems like not a big deal to me)

Also, what is the user "Cancelling" when touching "cancel"?

mohitsvnit:
You are right. I will remove cancel button.

ArihantJain456:
@mohitsvnit I think the proper alignment of labels is missing.
you can try something like this 
file dialog

mohitsvnit:
screenshot from 2018-04-03 14-06-41

screenshot from 2018-04-03 14-07-08

screenshot from 2018-04-03 14-13-37

@cpg which one do you suggest?

cpg:
definitely first one of the three, in my view.

i would also take away the file type line. who cares about it?
user probably has a good idea of what that file is anyway.

basically, i don't think this feature very important (not high priority), but if we do it, we should do it cleanly and without junk on it.

mohitsvnit:
Ok, I will remove file type. Do you want me to increase text size to `16sp` as suggested in material design guideline? Currently, it is `14sp`. 

The 3rd one is with `16sp` text size.

cpg:
i don't know. it's going to make some file names overflow. I would not mind if it's smaller.

actually, haha ...  honestly, ... without the file info (which is not really necessary), all that information is available in the entry for the file already, so this does not add much value, i think ...

mohitsvnit:
screenshot_1522755242

I updated the code as per everyone suggestions.

mohitsvnit:
screenshot_1522765963

Updated code as per everyone suggestions.

cpg:
OK. After more understanding of what this is an after a design analysis, I'm reticent to push for it. It's not really immediately useful to the user as it is.

The only way I can think this could be useful is if we move things from the nav header to this pop up. For example if we move the "delete the file" button, or the "share the file" button there. That would allow us to add more things over time, at the expense of one extra tap for each such thing

However, the thing is we _do not really need expansion yet._ So I would like to close this now but we can re-open it an use the work done once we need the expansion. Thanks for your patience with my questions an slow response @mohitsvnit 

Closes #342

Issue #300, opened 2018-03-02
Replacing the spinner in the navigation bar with a better material design approach


octacode:
### Present Scenario
The spinner in the navigation bar where the name of the server appears and from where we can switch between our local HDAs or to the `Welcome to Amahi Server` looks a bit wierd
![screenshot_20180302-143703](https://user-images.githubusercontent.com/20819968/36891377-6d7abf88-1e27-11e8-9427-806799512ec6.png)

We can replace it with just a text clicking on which shows up the AlertDialogBox which allows switching of the servers or maybe do something fancy like in case of the Gmail app.

mohitsvnit:
I would like to work on it

octacode:
Thanks for volunteering, Can you propose a solution for this before you start working on it?

mohitsvnit:
What is purpose of that spinner? In my mobile I can see only one item in it i.e. "Welcome to Amahi"

mohitsvnit:
Ok. I got it. It's a way to switch between server. Why I am not able to see my server? 
How about making it like gmail? We can use material design drawer design

adi23arora:
@octacode This can be implemented as an expandable layout with list view. It won't overlap then and expand within the navigation drawer. What do you say?

octacode:
Sounds like a plan to me, but won't the animation of expanding and contracting be a bit troublesome and add unneccesary overhead. My suggestion is that we do it the way we do it in the Gmail App:

1. We can add a header on the navigation bar with the name of the server like this
![header](https://user-images.githubusercontent.com/20819968/36895323-8726561a-1e34-11e8-84db-64200aea751a.png)
also below the header we'll have the `Shares` and the `Apps` like the current version.

2. On clicking the down arrow button a new list will replace the current list and this new list will contain the name of the HDAs, user can select the HDA from here. Just like this below
![two](https://user-images.githubusercontent.com/20819968/36895609-7f76fa04-1e35-11e8-9b0b-c59092d031d8.png)

cpg:
I don't really understand what's being propose, but it looks weird to have that large logo.

also the "Signed in as:" is a little superfluous.

octacode:
We can change the text to "HDA:" and about the sign we can decrease the height of header so that it doesn't look weird.

csoni111:
1) The ActionBar on the top already has the amahi logo on the left of the title. If we remove that from top, what about other screens?
2) The API currently does not provide the logged in user's name.
[...]

PR #356, opened 2018-03-27
Navigation Drawer Redesigned According to Material UI Guidelines

sanjitchak:
Made changes according to comments given in  #300 by @octacode  & @adi23arora while adhering to [Material UI GuideLines](https://material.io/guidelines/patterns/navigation-drawer.html#navigation-drawer-specs) too.


src/main/java/org/amahi/anywhere/activity/NavigationActivity.java
```diff
@@ -145,7 +145,9 @@ private void setUpInjections() {
     private void setUpHomeNavigation() {
         getSupportActionBar().setHomeButtonEnabled(isNavigationDrawerAvailable());
         getSupportActionBar().setDisplayHomeAsUpEnabled(isNavigationDrawerAvailable());
+/*
```
ArihantJain456: Whats the need for commenting instead of removing?
Also if you want to comment use //
Most text editors/ IDE provide commenting/ uncommenting feature using `ctrl` + `/`

src/main/java/org/amahi/anywhere/adapter/NavigationDrawerAdapter.java
```diff
@@ -1,32 +1,49 @@
...
 
 public class NavigationDrawerAdapter extends RecyclerView.Adapter<NavigationDrawerAdapter.NavigationDrawerViewHolder> {
 
-    private final List<Integer> navigationItems;
+    private final List<Integer> mNavigationItems;
     private Context mContext;
 
-    public NavigationDrawerAdapter(Context context, List<Integer> navigationItems) {
-        this.navigationItems = navigationItems;
+    private final ArrayList<String> mServerName;
```
ArihantJain456: Code not refactored

src/main/java/org/amahi/anywhere/adapter/NavigationDrawerAdapter.java
```diff
@@ -1,32 +1,49 @@
...
+
+
+    //for Old Navigation List
+    private NavigationDrawerAdapter(Context context, List<Integer> navigationItems) {
```
ArihantJain456: do we need it?

src/main/java/org/amahi/anywhere/adapter/NavigationDrawerAdapter.java
```diff
@@ -40,14 +57,44 @@ public NavigationDrawerViewHolder onCreateViewHolder(ViewGroup parent, int viewT
         return new NavigationDrawerViewHolder(LayoutInflater.from(parent.getContext()).inflate(R.layout.view_navigation_item, parent, false));
     }
 
+    private Drawable getServerBubble(ArrayList<String> serverList, int position) {
+        //get first letter of each String item
+        String firstLetter = String.valueOf(serverList.get(position).charAt(0));
+
+        ColorGenerator generator = ColorGenerator.MATERIAL; // or use DEFAULT
```
ArihantJain456: why random?

src/main/java/org/amahi/anywhere/adapter/NavigationDrawerAdapter.java
```diff
@@ -40,14 +57,44 @@ public NavigationDrawerViewHolder onCreateViewHolder(ViewGroup parent, int viewT
...
+        holder.server_bubble.setImageResource(R.drawable.ic_app_logo_shadowless);
+        holder.server_bubble.setScaleType(ImageView.ScaleType.CENTER_CROP);
+
+
     }
 
     @Override
     public int getItemCount() {
-        return navigationItems.size();
+        if (mNavigationItems != null) {
```
ArihantJain456: This may be better.
`return mNavigationItems!=null?mNavigationItems.size():mServerName.size();`
Also is there are cases when we are getting null list?

src/main/java/org/amahi/anywhere/fragment/NavigationFragment.java
```diff
@@ -111,6 +116,8 @@ public void onActivityCreated(Bundle savedInstanceState) {
         setUpContentRefreshing();
 
         setUpServers(savedInstanceState);
+
```
ArihantJain456: do proper code refactor.

src/main/java/org/amahi/anywhere/adapter/NavigationDrawerAdapter.java
sanjitchak: I understand your concern for the null list but here, we are changing the list items from the same area with either server Names or with selected server option (shares, apps).

src/main/java/org/amahi/anywhere/adapter/NavigationDrawerAdapter.java
sanjitchak: Array of String is for **server name** list. List of Integer is for **Selected Server Option (share,apps,etc)**, made by someone before.

src/main/java/org/amahi/anywhere/adapter/NavigationDrawerAdapter.java
adi23arora: I guess random colors are needed for that label thing corresponding to the server name. But there could be a different approach (like creating an array of colors for that), instead of adding a new library.

sanjitchak: Yep, true. I could have done that with an array of colors for different alphabets. 
And that library does pretty much everything, including making a circle of random color with an appropriate alphabet while being lightweight too.

adi23arora:
@sanjitchak Build failed due to missing drawable files `nav_arrow_down` and `nav_arrow_up`, add them. And remove the extra drawable files `quantum_ic_keyboard_arrow_up_white_36.png`, which aren't used anywhere.

sanjitchak:
@adi23arora Thanks! for the heads-up. Made necessary changes.

cpg:
I merged this just because it's claimed to be more compliant but I ~kind of~ hate it.

I made some changes: `Server Name:` should be `Server` and not in bold. If anything the actual name of the server should be in **bold**.

The following are changes I would suggest we do *on top* of this code:
* the small Amahi dice logos next to "Shares" and "Apps" are superfluous and overloading. Maybe a folder for `Shares` and some other one for `Apps`
* the name of the server disappears sometimes, just shows the literal string `Server Name` which is confusing. I did not change the server! -- I think this should be reproducible by rotating he phone.
* the Amahi dice logo inside the circle is truly abominable to look at. two corners touch the edge of the circle (!!). The word "Server" looks out of place.
* the flipping the arrow to select another server just looks flat wrong
* the circle with the first letter of the server name looks neat, etc., though the colors are very bright and seem out of place with the theme of the app.
* the first one (circle around the first letter of the server) is bright green and the letter inside is while, which makes it virtually impossible to see.

In writing this I now regret merging this one .... my bad. I think this is a regression in usability and looks.

adi23arora:
@cpg I just tested this PR, the flipping arrow switch doesn't seem to work and there are some other minor bugs too.
I also looked at your suggestions, these changes are necessary for good user experience.

sanjitchak:
@cpg Thanks for your valuable feedback. I highly appreciate it. 

I am also regretting myself for raising such a horrible PR, after reading your suggestions. While complying with material UI guidelines, I utterly failed to see the user experience perspective.

@adi23arora  So, would you mind allowing me to work on the suggestions if you are not already working on it?

cpg:
@sanjitchak Please go ahead and work on this as time permits. Let's do it under a separate issue #365 and a separate PR.

Issue #357, opened 2018-04-02
Foreground audio service not functioning properly


adi23arora:
### Issue Description
Due to some recent commits, for adding the swipe feature in PR #331, ``AudioService.java`` and other related files were modified which broke the foreground audio service feature.

While playing an audio file when the app is switched to background, the audio controls in media notification, for switching to previous and next song doesn't seem to work. Moreover, the song isn't switched to next song even on completion.

I have tested that these controls were working fine previously.

### Steps to Reproduce
1. Play an audio file from Amahi app, from any server.
2. Switch the app to background
3. Try to use the next and previous controls in media notification 
   (OR) wait for the song to finish

ArihantJain456:
Thanks @adi23arora for pointing it out, I will look into it.

ArihantJain456:
I know why they are not working, because I removed those methods next, previous and on completion in 'audioService.java'.

adi23arora:
@ArihantJain456 I checked the code in `AudioService.java` and found some of the methods we were using for handling those events were removed. 
And you were posting new events like `AudioControlNextEvent()` and `AudioControlPreviousEvent()`, without specifying any action on detecting those event.

adi23arora:
Can I try fixing this, I will seek your help if required.

ArihantJain456:
I have already fixed it. I have classes from 11:00 am to 6:pm, so I will send a PR after it. But if you want to try I will  not send my PR.

ArihantJain456:
Hey @adi23arora, you haven't answer that question ^. If you are willing to work on this then I will help you in this.

adi23arora:
@cpg This issue is resolved with PR #360, can be closed now.

PR #358, opened 2018-04-03
Fixed audio control from media notification

adi23arora:
Fixes https://github.com/amahi/android/issues/357

### DIFF
--- a/src/main/java/org/amahi/anywhere/service/AudioService.java
+++ b/src/main/java/org/amahi/anywhere/service/AudioService.java
@@ -311,7 +311,9 @@ public class AudioService extends MediaBrowserServiceCompat implements
 
     @Subscribe
     public void onAudioControlChange(AudioControlChangeEvent event) {
-        startChangedAudio(event.getPosition());
+        if (audioFile != getChangedAudioFile(event.getPosition())) {
+            startChangedAudio(event.getPosition());
+        }
     }
 
     @Subscribe
--- a/src/main/java/org/amahi/anywhere/service/AudioService.java
+++ b/src/main/java/org/amahi/anywhere/service/AudioService.java
@@ -274,6 +274,10 @@ public class AudioService extends MediaBrowserServiceCompat implements
         return audioFile;
     }
 
+    public int getAudioFilePosition() {
+        return audioFiles.indexOf(audioFile);
+    }
+
     public AudioMetadataFormatter getAudioMetadataFormatter() {
         return audioMetadataFormatter;
     }
@@ -308,6 +312,36 @@ public class AudioService extends MediaBrowserServiceCompat implements
     @Subscribe
     public void onAudioControlChange(AudioControlChangeEvent event) {
         startChangedAudio(event.getPosition());
+    }
+
+    @Subscribe
+    public void onAudioControlNext(AudioControlNextEvent event){
+        startChangedAudio(getNextAudioPosition());
+    }
+
+    @Subscribe
+    public void onAudioControlPrev(AudioControlPreviousEvent event){
+        startChangedAudio(getPreviousAudioPosition());
+    }
+
+    private int getNextAudioPosition() {
+        int currentPosition = getAudioFilePosition();
+
+        if (currentPosition == audioFiles.size()-1) {
+            return 0;
+        } else {
+            return currentPosition+1;
+        }
+    }
+
+    private int getPreviousAudioPosition() {
+        int currentPosition = getAudioFilePosition();
+
+        if (currentPosition == 0) {
+            return audioFiles.size()-1;
+        } else {
+            return currentPosition-1;
+        }
     }
 
     public void playAudio() {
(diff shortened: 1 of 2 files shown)


src/main/java/org/amahi/anywhere/service/AudioService.java
```diff
@@ -310,6 +314,36 @@ public void onAudioControlChange(AudioControlChangeEvent event) {
...
+
+    @Subscribe
+    public void onAudioControlPrev(AudioControlPreviousEvent event){
+        startChangedAudio(getPreviousAudioPosition());
+    }
+
+    private int getNextAudioPosition() {
+        int currentPosition = getAudioFilePosition();
+
+        if (currentPosition == audioFiles.size()-1) {
```
ArihantJain456: using conditional operator would be great.

adi23arora:
@ArihantJain456  I have fixed it completely now, please review.

ArihantJain456:
@adi23arora song is still not switching to next song on song completion (when the app is in background). You haven't changed `onCompletion` method in `AudioService`.

adi23arora:
Ok, I'll look into it.

ArihantJain456:
@adi23arora good work. Now it is working fine, but I have one suggestion:
AudioService and ServerFileAudioActivity both are receiving `AudioCompleteEvent` and both are trying to change song which may be unnecessary, If something can be done that prevent this then it would be great.

adi23arora:
@ArihantJain456 We can't eliminate any one of these because of following reasons:
* In `AudioService.java`, we are using it switching audio even when the app is moved to the background.
* And in `ServerAudioFileActivity.java`, it is used for switching the view pager to update it to the current item.

ArihantJain456:
In the `ServerAudioFileActivity.java` we are using it for both changing songs and updating view pager.
I think we should update UI (view pager in Activity) and change song in Service.
@adi23arora what do you say?

ArihantJain456:
### The present scenario on completion of a song is as follows:
#### If the app is in background
- changing to the next song from audio service
- and play
#### If the app is not in background
- changing to the next song from service
- song changed but does not play due to some error
- updating view pager from Activity
- changing song from Activity by calling audio service
- this time song plays

### What I was suggesting
####  If the app is in background
- change song from service 
- and play
#### If the app is not in background
- update  view pager from Activity
- change song from service 
- and play

src/main/java/org/amahi/anywhere/service/AudioService.java
```diff
@@ -316,17 +316,17 @@ public void onAudioControlChange(AudioControlChangeEvent event) {
...
     @Subscribe
     public void onAudioControlPrev(AudioControlPreviousEvent event) {
-        startChangedAudio(getPreviousAudioPosition());
+        BusProvider.getBus().post(new AudioControlChangeEvent(getPreviousAudioPosition()));
     }
 
     @Subscribe
     public void onAudioCompleted(AudioCompletedEvent event) {
-        startChangedAudio(getNextAudioPosition());
+        BusProvider.getBus().post(new AudioControlChangeEvent(getNextAudioPosition()));
```
ArihantJain456: This creates one additional call. So now, the audio player is preparing for the next position three times.

src/main/java/org/amahi/anywhere/service/AudioService.java
```diff
@@ -316,17 +316,17 @@ public void onAudioControlChange(AudioControlChangeEvent event) {
 
     @Subscribe
     public void onAudioControlNext(AudioControlNextEvent event) {
-        startChangedAudio(getNextAudioPosition());
+        BusProvider.getBus().post(new AudioControlChangeEvent(getNextAudioPosition()));
```
ArihantJain456: Audio Player is now preparing two times.

src/main/java/org/amahi/anywhere/service/AudioService.java
```diff
@@ -316,17 +316,17 @@ public void onAudioControlChange(AudioControlChangeEvent event) {
...
     }
 
     @Subscribe
     public void onAudioControlPrev(AudioControlPreviousEvent event) {
-        startChangedAudio(getPreviousAudioPosition());
+        BusProvider.getBus().post(new AudioControlChangeEvent(getPreviousAudioPosition()));
```
ArihantJain456: similar to[ line#319](https://github.com/amahi/android/pull/358/commits/98bd6cdb02e5eabde08d507f2ca68ad623f8fd2a#diff-73396d7ae4339ff7848b0e2e8d1cfdf1R319)

ArihantJain456:
When I was testing the notification I have noticed one more issue related to the next and previous button.
Steps to reproduce:
1. click on next button in the notification
2. click on the notification to go to the audio player activity before player prepares.
3. And now the player will not play the song. <- this is the issue.

No need to rush this time.

cpg:
Ok, sorry I had to merge the larger one and that produced conflicts. Please take a look to fix the conflicts.

adi23arora:
@cpg I guess, work on this PR is no longer needed as the changes targeted in this were already resolved with Exoplayer implementation in PR #360.
I continued working on this because as discussed with you on IRC, I just thought exoplayer is experimental. But now when it is merged, should I close this one?

PR #360, opened 2018-04-11
Replacing MediaPlayer with Exoplayer for playing audio files

### LINKED ISSUE #165: Use ExoPlayer in place of MediaPlayer(MediaController)
About Exoplayer :
The Android framework provides MediaPlayer as a quick solution for playing media with minimal code, and the MediaCodec and MediaExtractor classes are provided for building custom media players. ExoPlayer is an open source project that provides a solution falling somewhere between these two options. It is a player that you can modify and extend.
 Also it includes tunneled video playback.
Primarily intended for Android TVs, tunneled video playback promises benefits such as better audio/video synchronization (AV sync) and smoother playback. 
https://developer.android.com/guide/topics/media/exoplayer.html

ArihantJain456:
This includes following changes:
- Implementation of ExoPlayer
- Minor bug fixes in Audio Player

closes #165

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -145,7 +145,7 @@ dependencies {
     androidTestImplementation 'com.android.support:support-annotations:25.4.0'
     compileOnly 'com.squareup.dagger:dagger-compiler:1.2.5'
     implementation 'com.github.wseemann:FFmpegMediaMetadataRetriever:1.0.14'
-
+    implementation 'com.google.android.exoplayer:exoplayer:2.7.3'
     // Uncomment the dependencies below to enable Chuck Interceptor for logging
 /*
     debugImplementation 'com.readystatesoftware.chuck:library:1.1.0'
--- a/build.gradle
+++ b/build.gradle
@@ -36,7 +36,9 @@ android {
         }
     }
 
-
+    dexOptions {
+        jumboMode true
+    }
 
     defaultConfig {
         def versionMajor = 3
@@ -145,7 +147,8 @@ dependencies {
     androidTestImplementation 'com.android.support:support-annotations:25.4.0'
     compileOnly 'com.squareup.dagger:dagger-compiler:1.2.5'
     implementation 'com.github.wseemann:FFmpegMediaMetadataRetriever:1.0.14'
-    implementation 'com.google.android.exoplayer:exoplayer:2.7.3'
+    implementation 'com.google.android.exoplayer:exoplayer-core:2.7.3'
+
     // Uncomment the dependencies below to enable Chuck Interceptor for logging
 /*
     debugImplementation 'com.readystatesoftware.chuck:library:1.1.0'
--- a/src/main/java/org/amahi/anywhere/service/AudioService.java
+++ b/src/main/java/org/amahi/anywhere/service/AudioService.java
@@ -42,6 +42,7 @@ import android.support.v4.media.session.PlaybackStateCompat;
 import android.text.TextUtils;
 
 import com.google.android.exoplayer2.DefaultLoadControl;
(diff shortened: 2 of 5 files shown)


src/main/java/org/amahi/anywhere/service/AudioService.java
```diff
@@ -144,15 +153,18 @@ private void setUpBus() {
...
+        audioPlayer.setPlayWhenReady(true);
+        audioPlayer.setVolume(1.0f);
+    }
 
-        audioPlayer.setOnPreparedListener(this);
-        audioPlayer.setOnCompletionListener(this);
-        audioPlayer.setOnErrorListener(this);
+    private MediaSource buildMediaSource(Uri uri) {
+        return new ExtractorMediaSource.Factory(
+            new DefaultHttpDataSourceFactory(getString(R.string.application_name)))
```
csoni111: Instead of using `application_name` as user-agent, you can use `Identifier.getUserAgent(context)`.

src/main/java/org/amahi/anywhere/service/AudioService.java
```diff
@@ -144,15 +153,18 @@ private void setUpBus() {
     }
 
     private void setUpAudioPlayer() {
-        audioPlayer = new MediaPlayer();
-
-        audioPlayer.setAudioStreamType(AudioManager.STREAM_MUSIC);
-        audioPlayer.setWakeMode(this, PowerManager.PARTIAL_WAKE_LOCK);
-        audioPlayer.setVolume(1.0f, 1.0f);
+        audioPlayer =
+            ExoPlayerFactory.newSimpleInstance(
```
csoni111: `newSimpleInstance(Context, TrackSelector, LoadControl)` is deprecated, you may use `newSimpleInstance(RenderersFactory, TrackSelector, LoadControl)`.

build.gradle
```diff
@@ -145,7 +145,7 @@ dependencies {
     androidTestImplementation 'com.android.support:support-annotations:25.4.0'
     compileOnly 'com.squareup.dagger:dagger-compiler:1.2.5'
     implementation 'com.github.wseemann:FFmpegMediaMetadataRetriever:1.0.14'
-
+    implementation 'com.google.android.exoplayer:exoplayer:2.7.3'
```
csoni111: You have included the full exoplayer library. This could be the reason for the build failing on travis, as the max strings allowed in dex is exceeding than the defined limit. As per your code looks like you require only core functionality.
```
implementation 'com.google.android.exoplayer:exoplayer-core:2.7.3'
```

ArihantJain456: I have tried it. But still, it is giving the same error. Using 
`dexOptions {
        jumboMode true
    }`
seems to solve the problem.

csoni111: That was the last resort. Anyway, still reduce the dependency to `core` only (might save us some apk size).

ArihantJain456: ok, I will change the dependency to include exoplayer-core. So, using `dexOptions { jumboMode true }` would be fine?

csoni111: yes

cpg:
Testing this for a few hours. So far so good. Great job.

When a song is playing and I rotate the phone, I get a spinner there in the middle of the screen ... until the song ends, at which time the song artwork comes up properly and the spinner goes away. Looks like it needs a handler for rotation events.

That said, I _think_ I reported (or at least I saw this) before, so I am not sure this is related to this particular PR.

cpg:
Great job!

adi23arora:
@cpg I thought it was supposed to be merged in `exoplayer` branch.

PR #368, opened 2018-05-03
adds download option and offline mode

### LINKED ISSUE #355: No direct Download option
## Description
For videos, audios, and images there is no direct option for downloading.
If a user wants to download a file then only way around is to click on share option.

### Solution 
 Download option can be added to the action bar along with other options.

ArihantJain456:
- [x] Replace action mode with bottom sheet dialog
- [x] Implement download option
- [x] Changing database to accommodate offline mode
- [x] Adding Separate `Offline Share`

closes #355


cpg:
maybe it's best to get the state diagram in separately in a different PR.
let's settle on the states/transitions as we see fit and then we work to get it working.

adi23arora:
@ArihantJain456 You have used `Room` library for database implementation, but due to size concerns here https://github.com/amahi/android/issues/311#issuecomment-370790225 @octacode suggested to use SQLite APIs only.

ArihantJain456:
I compared the size of the latest apk file and the apk file without the room, the difference is about 40 KB.
Are you are talking about this? @adi23arora

adi23arora:
Great, then. `Room` indeed is the best option for databases.

cpg:
Great job! Feedback:

1) I tried downloading one file, and it appeared to work. I did not click on open right away. I went to the offline files area and there were no files there. There was only the empty message.

Then I realized that there is a wording issue. When downloading, it does not mean that a file will be offline. Because of the fact that the we have two unrelated things using "download" or similar, we will have to change some of the strings.

2) When I access the offline files area, I see there is an option to Share, one to Download and one to Delete. The Download option is confusing also. What does it do?

ArihantJain456:
- `download` option is for downloading the file in the public download directory of the phone and once downloaded, these files will not have any connection to the app. These files are accessible by all apps.
- the `offline` option is for downloading Amahi app supported file (more likely media files) in the internal storage of the phone (accessible through Amahi app only).

PR #371, opened 2018-05-21
Fixes Uri parsing issue (#350)

### LINKED ISSUE #350: Can't share music and video files
## Description
Error occurs on sharing video and audio files.

### Steps to reproduce
1. long press any video/audio file 
2. select share option
3. Wait for the file to download
4. Select any sharing app.
5. See the  error

### Updates 
After further testing, I found that the problem is with **Welcome to Amahi** server's video/audio files only.
Sharing feature is working fine with my remote and Local server.

@octacode please give some suggestions on this.

ArihantJain456:
fixes #350.

### DIFF
--- a/src/main/java/org/amahi/anywhere/util/Downloader.java
+++ b/src/main/java/org/amahi/anywhere/util/Downloader.java
@@ -28,6 +28,7 @@ import android.database.Cursor;
 import android.net.Uri;
 import android.os.Environment;
 import android.support.v4.content.FileProvider;
+import android.util.Log;
 
 import org.amahi.anywhere.bus.BusProvider;
 import org.amahi.anywhere.bus.FileDownloadFailedEvent;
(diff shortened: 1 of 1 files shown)


src/main/java/org/amahi/anywhere/util/Downloader.java
```diff
@@ -115,12 +117,15 @@ private void finishDownloading() {
             String downloadUri = downloadInformation.getString(
                 downloadInformation.getColumnIndex(DownloadManager.COLUMN_LOCAL_URI));
 
-            if (downloadUri.substring(0, 7).matches("file://")) {
-                downloadUri = downloadUri.substring(7);
+            try {
+                URI uri = new URI(downloadUri);
+                downloadUri = uri.getPath();
+            } catch (URISyntaxException e) {
+                e.printStackTrace();
```
cpg: minor niggle. wouldn't we want to print the URI, tho, not the stack trace?

ArihantJain456: Stack trace contains the URI, but I too feel that printing URI only will work fine plus there is one more change that I think is necessary (`/file:` need to be removed from the Uri), so I will update the PR.

Issue #377, opened 2018-06-10
Offline folder is not accessible in offline mode


codedentwickler:
When the app is launched with no internet connectivity or in airplane mode, the offline folder is not accessible, only an error message "Connection Error" shown in the Navigation Drawer. This defeats the purpose of making the files available offline in the first instance.

We can replace this error message with a text indicator maybe `Currently Offline` and show the offline folder in the Navigation Drawer.

ArihantJain456:
We can solve it this way:
1. If no server is selected, this means the user is opening the app for the first time. If there is no internet connection, we can show the same network error.
2. If a server is selected before. We can show the selected server and offline files option, and a text saying offline.

njnaman:
if this issue is open can i work in this?

chirag-jn:
This issue doesn't exist anymore.

PR #380, opened 2018-06-13
changes in navigation drawer

### LINKED ISSUE #365: navigation drawer improvements
The following are some changes I would suggest we do for the nav drawer:

* the small Amahi dice logos next to "Shares" and "Apps" are superfluous and overloading. Maybe a folder for `Shares` and some other one for `Apps`
* the name of the server disappears sometimes, just shows the literal string `Server Name` which is confusing. I did not change the server! -- I think this should be reproducible by rotating he phone.
* the Amahi dice logo inside the circle is truly abominable to look at. two corners touch the edge of the circle (!!). The word "Server" looks out of place.
* the flipping the arrow to select another server just looks flat wrong
* the circle with the first letter of the server name looks neat, etc., though the colors are very bright and seem out of place with the theme of the app.
* the first one (circle around the first letter of the server) is bright green and the letter inside is while, which makes it virtually impossible to see.

ArihantJain456:
Changes in the navigation drawer:
1. change in icons of shares and apps menu
2. changes color of server icons to primary color of the app.
3. fixes flipping arrow 
4. makes the first server as the default server in case no server is selected.

Changes remaining:
- [x] changes in navigation drawer header
- [x] full height drawer

 

closes #365 #300 #377


octacode:
@cpg Does this drawer looks fine to you?

cpg:
I thought we decided to call them "Downloaded" rather than "Offline files" ?

PR #387, opened 2018-07-02
adds audio navigation

ArihantJain456:
- [x] feature to navigate through a share while playing music
- [x] removed the feature to play audio from the last played position
- [x] adds audio files list in the audio player activity


octacode:
Hi, I did test the music player feature. It looks fine to me. However, I was thinking of a button which shows the list of songs which are to be played after the current song ends. Like a current playing playlist
You can look at Google play music, VLC player or any other music player for the workflow.

ArihantJain456:
We can add a menu to the toolbar to access the list of songs just like VLC Google play music.

PR #388, opened 2018-07-14
adds NAUs Access

ArihantJain456:
- [x] add pin login for local servers
- [x] add HDA cache
- [x] change files access API for downloading and streaming files.
- [x] handling share write permission


csoni111:
App crashes on going into settings -> changing connection mode to Remote with this in logs:
```
java.lang.NullPointerException: baseUrl == null
        at retrofit2.Utils.checkNotNull(Utils.java:286)
        at retrofit2.Retrofit$Builder.baseUrl(Retrofit.java:451)
        at org.amahi.anywhere.server.ApiAdapter.create(ApiAdapter.java:50)
        at org.amahi.anywhere.server.client.ServerClient.buildServerApi(ServerClient.java:142)
        at org.amahi.anywhere.server.client.ServerClient.connectRemote(ServerClient.java:239)
        at org.amahi.anywhere.fragment.SettingsFragment.setUpServerConnection(SettingsFragment.java:273)
        at org.amahi.anywhere.fragment.SettingsFragment.onSharedPreferenceChanged(SettingsFragment.java:254)
```

src/main/java/org/amahi/anywhere/server/api/ServerApi.java
```diff
@@ -46,32 +46,27 @@
 public interface ServerApi {
     @POST("/auth")
     Call<HdaAuthResponse> authenticate(
-        @Header("Session") String session,
```
csoni111: Why did you remove the session header? The remote browsing has stopped working.
The session token is for the pfe to identify which fs to connect to (so its not needed in local mode but required in remote mode) where as the auth token is for the fs to identify which hda-user to serve the content to.

csoni111: You may keep the session header as optional, when you have it in case of remote browsing, send it. When in local mode (using pin access), don't send it as you don't have it.

ArihantJain456: ok, I will restore the session code. For local pin access what there is no remote address, Should I hide remote connection from settings for it or set `remote address = local address`?

csoni111: For now hide/disable the remote option (if its not too difficult to do so  or else make equal). In future we have to implement remote access for NAU also, so mark a TODO in code where you make this change.

csoni111:
These are few a points I noted down while testing:
1) If there are multiple HDAs available for a user, it asks PIN only for the first HDA selected and for others it uses the same _auth_token_. This should not be the case, each HDA will have different _auth_token_ and different PIN.
2) The HDA name is not changing in side nav bar below the text `Server`, on selecting another HDA from the list.
3) On logging in via PIN access, if you go to `Auto Upload Photos` settings, it sends a request to `/servers` on main amahi api to fetch hda list with the logged in hda's auth token. The _access_token_ and _auth_token_ are different and should not be mixed.
4) When creating a new server via PIN access, the _session_token_ is set same as the _auth_token_. This won't make any difference as we are always connecting locally but this will need to be updated when remote connection is added for NAU.
5) When an api returns `401 Unathorized` -> it should reset the _auth_token_ previously stored (if any) and ask for the PIN again from the user instead of simply showing connection error. 401 means user is unauthenticated or the token expired. 
6) Similarly `403 Forbidden` means the user does not have permission to access that resource so accordingly error should be shown, something like, `You do not have permission, contact your administrator`. By the way this should never happen as the api already limits access at every step.

Apart from these, the app is working fine for NAU access as well as the PIN functionality for admin users.

ArihantJain456:
1. When a user selects another hda, it is asking for the selected hda's PIN and changes auth token on successful authentication. For the `Welcome to Amahi` server, it is not asking for the PIN.
3. If the user is logged in through PIN access, then it is asking for servers using auth_token otherwise working ok. For PIN access, fixing upload server to local server will work fine.

For the rest of the points, I will make improvements.

csoni111:
To reproduce the bug I mentioned in 1st point, follow these steps:
1) Add two custom servers in `assets/customServers.json`
	```json
	[
	    {
	        "name": "Test Server 1",
	        "session_token": "abcd",
	        "local_address": "http://192.168.0.104:4563",
	        "remote_address": "http://192.168.0.104:4563"
	    },
	    {
	        "name": "Test Server 2",
	        "session_token": "abcd",
	        "local_address": "http://192.168.0.104:4563",
	        "remote_address": "http://192.168.0.104:4563"
	    }
	]
	```
2) Login into the app as admin user
3) Select `Test Server 1` -> enter the pin -> display shares.
4) Now select `Test Server 2` -> it doesn't ask you for pin again, it simply sends the request with `Test Server 1`'s _auth_token_.
5) Also the name of the server selected is not changed in the side nav bar.

ArihantJain456:
Well, two servers cannot have the same session, right? on selecting another server, PIN Access screen will be visible only when servers are different (their session are different)? If that's not ok, we can show PIN access screen when servers names are different just as above (`Test Server 1` & `Test Server 2`) ?

csoni111:
Ya this is correct. My mistake, I did not take care of the same _session_token_.

ArihantJain456:
@csoni111 I have fixed all the above issues. Please review it when you have a free moment.

octacode:
I have verified the ATV changes, Thanks for reviewing it @csoni111

PR #390, opened 2018-10-25
Add french strings

Eradash:
There is some missing translations, the context was not clear enough for me. I did not build the application, try it to let Android Studio spot them (surely some ' without \ before)

### DIFF
--- a/src/main/res/values-fr/strings.xml
+++ b/src/main/res/values-fr/strings.xml
@@ -1,0 +1,171 @@
+<?xml version="1.0" encoding="utf-8"?><!--
+  ~ Copyright (c) 2014 Amahi
+  ~
+  ~ This file is part of Amahi.
+  ~
+  ~ Amahi is free software: you can redistribute it and/or modify
+  ~ it under the terms of the GNU General Public License as published by
+  ~ the Free Software Foundation, either version 3 of the License, or
+  ~ (at your option) any later version.
+  ~
+  ~ Amahi is distributed in the hope that it will be useful,
+  ~ but WITHOUT ANY WARRANTY; without even the implied warranty of
+  ~ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+  ~ GNU General Public License for more details.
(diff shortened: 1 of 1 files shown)


cpg:
It looks like it failed to build. Did you have a chance to built it yourself?

Eradash:
It looks like the failed build is not due to my modifications...
image
I tried on my computer, and it's the same result for my branch, and the `beta` branch.
Even when updating all tools and dependencies, the project would not build... From what I know, the recyclerview-v7 is included multiple times in dependencies, and must be excluded to keep only one

cpg:
@ArihantJain456 that being the case, it's really up to us to remove that, not up to him providing an unrelated PR, really. No?

Eradash:
Even after removing the dependency, there is still the same error appearing on my side

ArihantJain456:
@Eradash the error is still occuring since it is trying to download recyclerview from bintray first instead of maven, which I have taken care in the PR #391.

ArihantJain456:
@cpg can you rerun the Travis build?

cpg:
It passed, though I did notice this:

```
warning: string 'application_name' marked untranslatable but exists in locale 'fr'
```

cpg:
does the warning mean that there are other things that need to be done in the code to enable the `fr` strings? or is adding the strings file for `fr` enough?

is it automatic based on the user's language in the OS or does it require user selection or override?

Eradash:
The warning appears because the string is present in the french ones, even if the line was marked as not translatable. I simply forgot to remove the line...

Only adding the file named exactly as the english one is enough. It is placed in a directory tagged `-fr` to indicate the OS to look for this file first if the locale is set to french. It is automatic.

cpg:
Thanks for the explanation. Are you saying there is one extra line that needs to be removed now?

Eradash:
`<string name="application_name" translatable="false">Amahi</string>`
This line in the French file is not necessary since it's defined as non translatable in the main file. I did not check the tags, only the values...

PR #407, opened 2019-01-24
Created issue template

gopikaharikumar:
Created an issue template.


cpg:
GH has a template for all this, it turns out.

PR #404, opened 2019-01-22
Updated target and compile SDK version

RajaVamsi11:
Fixes #402 
Updated target and compile SDK versions of the app in build.gradle file

PR #415, opened 2019-02-21
fix:Updated target and compile SDK version

gopikaharikumar:
Fixes #404


cpg:
Can you clarify how is this any different than what's in #404?

PR #416, opened 2019-02-25
Replaced RelativeLayout with LinearLayout to fix android:layout_weight error

saarthak08:
Added contentDescription for all ImageViews & Replaced Relative Layout with Linear Layout in view_server_file_item.xml.

### DIFF
--- a/src/main/res/layout-land/layout_authentication.xml
+++ b/src/main/res/layout-land/layout_authentication.xml
@@ -21,12 +21,14 @@
     xmlns:app="http://schemas.android.com/apk/res-auto"
     android:layout_width="match_parent"
     android:layout_height="match_parent"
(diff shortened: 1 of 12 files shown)


saarthak08:
@cpg Can you please review this PR?

cpg:
What does this actually do, as in ... visually, functionally or otherwise?

saarthak08:
I removed all the errors & warnings in the XML layouts of all the files plus in view_server_file_item.xml, at one place there was Relative Layout used with a child attribute of layout:weight. Layout:weight attribute can't be used with relative layout & thats why it was showing error there. I removed the error by replacing it with LinearLayout. There were no changes in the view, visually it remained same but now, views rendering would perform/run better due to all errors & warnings removed.

PR #419, opened 2019-02-28
Dialog box for deletion confirmation

chirag-jn:
Dialog box added to confirm before deletion in offline files to prevent any deletion by mistake.


chirag-jn:
The app is successfully running after the changes made. Maybe there's some other issue. I'll look into it.

src/main/java/org/amahi/anywhere/activity/OfflineFilesActivity.java
```diff
@@ -180,7 +182,23 @@ private View getParentView() {
 
     @Subscribe
     public void onFileDeleting(OfflineFileDeleteEvent event) {
-        deleteFileFromOfflineStorage(event.getFile());
+        AlertDialog.Builder builder = new AlertDialog.Builder(this);
+        builder.setTitle("Delete file");
```
octacode: I would recommend putting these strings visible to the user in strings.xml

chirag-jn: Okay, thanks. I will do that.

codedentwickler:
@octacode @cpg I think this PR can be merged now, all the proposed changes has been completed. We have too many open PRs.

chirag-jn:
@cpg I guess we can merge this Pr now. Its review is completed a long time back.

Issue #411, opened 2019-02-02
It should Show an empty view when there's nothing to show on the Home Page


iadeelzafar:
I made an account on Amahi's website and then when I logged in using the mobile app it shows nothing when there are no HDA files on the Home Page. 

<img src = "https://user-images.githubusercontent.com/36811908/52166320-8a5ef000-272d-11e9-817e-6fd63ca7b09c.jpg" height="550" width="329">

It will be much better if it shows an empty view. An empty view to mention something like "You haven't added any HDA file yet"
For example:
This is just an example. I'll make a decent one.

<img src="https://user-images.githubusercontent.com/36811908/52124584-7a6edf80-264b-11e9-816b-30aa4e5de1f5.png" height="550" width="329" >

**I'd love to work on this. Just let me know if I can start working on this.**

iadeelzafar:
I'd love to work on this. Just let me know if I can start working on this.

Sreyas-108:
Its been more than 24 days since this issue has been raised. Can I work on this issue as my first contribution?

Sreyas-108:
Its been more than 24 days since this issue has been raised. Can I work on this issue as my first contribution? @cpg

cpg:
@iadeelzafar for something relatively simple like this there is no need to ask for permission and just wait. Best is to do it and submit a PR for review/merging. We were busy.

@Sreyas-108 ... in deference to @iadeelzafar, we should let him have a chance to work on it if he is still interested. If he does not reply in a few hours/days, we can go ahead.

iadeelzafar:
@Sreyas-108 Go ahead. Good Luck!

chirag-jn:
In accordance with #449, we are now showing shares on the home page. So we don't have to show an empty state screen anymore.

cpg:
We don't need this any more because we show shares.

PR #424, opened 2019-03-01
Empty state when there is nothing to show on the Home Page

Sreyas-108:
In reference to issue #411


iadeelzafar:
@Sreyas-108 whenever you're making a PR that has some UI changes. Don't forget to add the GIF showing those changes.

octacode:
Hi @Sreyas-108, it's a rarity that the user will will not have any shares. However, this addition won't do us any harm. There is also a very fair chance that the folders inside the shares are also empty. Are we displaying the empty message there? If not then that would be a good addition.

cpg:
@iadeelzafar maybe you can follow the suggestion?

Sreyas-108:
@cpg I am underway with the suggestion provided by @octacode . Will send a PR soon.

Sreyas-108:
@cpg I am underway with the suggestion provided by @octacode . Will send a PR soon.

iadeelzafar (changes requested):
Could you please squash your commits into a single one?

iadeelzafar (changes requested):
Also, could you please rename the commit message? Here's a useful resource to help you with commit messages. [How to write better commit messages](https://github.com/erlang/otp/wiki/writing-good-commit-messages)
You can do so by amending into your previous commit and then renaming.
Using ```git commit --amend```

Sreyas-108:
I will amend them accordingly. I havent squashed them since all changes havent been committed. Will do it once all changes are complete.

chirag-jn:
In accordance with #449, we are now showing shares on the home page. So we don't have to show an empty state screen anymore.

cpg:
We don't need this any more because we show shares.

PR #426, opened 2019-03-01
update Gradle version

zedlabs:
Gradle plugin does not work with the specified(old) Gradle version in android studio


nitinjangir:
notithl

PR #432, opened 2019-03-10
Removed the hard coded strings

### LINKED ISSUE #430: Removal of hard coded strings
**Summary**
The hard coded strings should be removed.

**Device and android version**
Device: Vivo V11 pro
Android version: 8.1.0

gopikaharikumar:
Fixes #430


gopikaharikumar:
@cpg can you please review this PR?

src/main/java/org/amahi/anywhere/activity/OfflineFilesActivity.java
```diff
@@ -168,7 +168,7 @@ public void onFileCopied(FileCopiedEvent event) {
     }
 
     private void dismissPreparingDialog() {
-        PrepareDialogFragment fragment = (PrepareDialogFragment) getSupportFragmentManager().findFragmentByTag("prepare_dialog");
+        PrepareDialogFragment fragment = (PrepareDialogFragment) getSupportFragmentManager().findFragmentByTag(getString(R.string.prepare_dialog));
```
codedentwickler: I dont see any reason for adding this to string.xml. It is not really a string resource, it is more of an id. We wouldn't want this to change if we decide to translate the App strings later. A better way would be to move it to the Constants class.

src/main/java/org/amahi/anywhere/activity/RecentFilesActivity.java
```diff
@@ -117,7 +117,7 @@ private void showIntroductoryOverlay() {
             new Handler().post(() -> {
                 mIntroductoryOverlay = new IntroductoryOverlay
                     .Builder(this, mediaRouteMenuItem)
-                    .setTitleText("Introducing Cast")
+                    .setTitleText(R.string.introducing_cast)
```
codedentwickler: Awesome.

src/main/java/org/amahi/anywhere/activity/RecentFilesActivity.java
```diff
@@ -394,7 +394,7 @@ private void showFileDownloadingFragment(ServerFile file) {
     }
 
     private void dismissPreparingDialog() {
-        PrepareDialogFragment fragment = (PrepareDialogFragment) getSupportFragmentManager().findFragmentByTag("prepare_dialog");
+        PrepareDialogFragment fragment = (PrepareDialogFragment) getSupportFragmentManager().findFragmentByTag(getString(R.string.preparedialog));
```
codedentwickler: Same here, better in the constant class.

src/main/java/org/amahi/anywhere/activity/ServerFilesActivity.java
```diff
@@ -389,7 +389,7 @@ private void openCamera() {
 
     private File createImageFile() throws IOException {
         String timeStamp = String.valueOf(new Date().getTime());
-        String imageFileName = "photo-" + timeStamp;
+        String imageFileName = getString(R.string.photo_) + timeStamp;
```
codedentwickler: Same here.

src/main/java/org/amahi/anywhere/fragment/SettingsFragment.java
```diff
@@ -112,7 +112,7 @@ private String getApplicationVersionSummary() {
     }
 
     private String getAutoUploadSummary() {
-        return isUploadEnabled() ? "Enabled" : "Disabled";
+        return isUploadEnabled() ? getString(R.string.enabled) : getString(R.string.disabled);
```
codedentwickler: Nice

src/main/java/org/amahi/anywhere/service/UploadService.java
```diff
@@ -219,7 +219,7 @@ private void setUpUploadManager() {
 
     private String queryImagePath(Uri imageUri) {
         String filePath = null;
-        if ("content".equals(imageUri.getScheme())) {
+        if (getString(R.string.content).equals(imageUri.getScheme())) {
```
codedentwickler: Same here.

codedentwickler (changes requested):
Good work overall. Move strings that should always remain the same (i.e ids or tags)to a Constants class. If this class do not exists, you can create one.

cpg:
@gopikaharikumar please take a look at the recommendations and see to either merge them or suggest what to do about it.

src/main/res/values/strings.xml
```diff
@@ -169,10 +169,35 @@
...
-
+    <string name="connection">CONNECTION</string>
+    <string name="shares">shares</string>
+    <string name="unknown_artist">Unknown Artist</string>
+    <string name="android_amahi_anywhere">Android Amahi Anywhere</string>
+    <string name="image_mediatype">image/*</string>
+    <string name="md5">MD5</string>
+    <string name="prepare_dialog_2">prepare_dialog</string>
+    <string name="progressdialog">progress_dialog</string>
+    <string name="prepare_dialog_3">prepare_dialog</string>
```
cpg, line 182: Is this and others (identical!) strings that the users see?

What should it be @octacode?

codedentwickler: They are not seen by the user. Putting them in strings.xml means they would be translated if we choose to localize the Amahi app in the future.

codedentwickler:
@gopikaharikumar As I said earlier, moving string constants the user do not see to strings.xml is not appropriate. Moving them to a Constants class where it remains the same even if we translate later is safer. 

This PR fixes a minor issue, please make the changes so we can merge it.

sumit-kushwah:
Can I work on this ?

chirag-jn:
Instead of adding constant strings to a Constants class, we can also add them to strings.xml and keep an attribute `translable="false"` in strings.xml itself.

cpg:
Fixed by #539. Good work!

Issue #433, opened 2019-03-10
getLastPathSegment() doesn't work with '?'


stupidly-logical:
**Describe the issue**
Using getLastPathSegment() does not get the complete file name from the URI. This occurs on this line: https://github.com/amahi/android/blob/dda04f1e6e5e5a4723e02e5134e457ecb5cdb653/src/main/java/org/amahi/anywhere/adapter/RecentFilesAdapter.java#L61
and also in RecentFile.java

It creates problem when getting the mime type because for file //special-chars/test ?.png it only gets 'test ' as file name.

**To Reproduce or Feature details**
1. Go to 'Shares>Demo>special-chars'
2. Click on 'test ?.png'
3. Go to Recent Files
4. App crashes

**Expected behavior**
Recent files should be populated

**Log of the value returned from getLastPathSegment() on the path '//special-chars/test ?.png' **
2019-03-10 16:52:00.992 24769-24769/org.amahi.anywhere D/LOG_MIME: RECENTFILE [test ]

**Smartphone (please complete as much of the following information as possible):**
 - Xiaomi Poco F1 - Android 9, API 28

**Additional context**
This may be because the function considers '?' as a terminal symbol.

PR #434, opened 2019-03-10
Fix getLastPathSegment() issue #443

stupidly-logical:
Explicitly using '/' to divide path into segments and getting the last segment which gives the files name. 
Fixes issue #433


stupidly-logical:
@cpg / @codedentwickler can you please review it.

cpg:
Great catch. Good job!

PR #442, opened 2019-03-18
Feature: Sort files by descending order of size

### LINKED ISSUE #441: Sort files by descending order of size
**Describe the issue**
Some times user wants to know which files have more size and viceversa. So letting the user sort according to descending order of file size will be really helpful.

I am going to work on this.

balaji-ramavathu:
Fixes #441 

**Screenshots**

sort-size

**Summary of the changes made**

Added sort_size icon and the flow of icons goes like this : sort_name - > sort_modification_time - > sort_size - > sort_name 
A comparator class is added to ServerFilesFragment to compare sizes of files in the list and sort them in descending order.

### DIFF
--- a/src/main/java/org/amahi/anywhere/fragment/ServerFilesFragment.java
+++ b/src/main/java/org/amahi/anywhere/fragment/ServerFilesFragment.java
@@ -119,6 +119,7 @@ public class ServerFilesFragment extends Fragment implements
 
     public static final int SORT_MODIFICATION_TIME = 0;
     public static final int SORT_NAME = 1;
+    public static final int SORT_SIZE = 2;
 
     @Inject
     ServerClient serverClient;
@@ -643,6 +644,9 @@ public class ServerFilesFragment extends Fragment implements
             case SORT_MODIFICATION_TIME:
                 return new FileModificationTimeComparator();
 
+            case SORT_SIZE:
+                return new FileSizeComparator();
+
             default:
                 return null;
         }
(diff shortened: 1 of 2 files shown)


csoni111 (commented):
See if you find a single vector image for the size sort icon instead of the png @balaji-ramavathu. Rest everything looks fine.

adi23arora:
When we have more than 2 sorting options, shouldn't that be displayed in a list from where the user can select the sorting option he wants, instead of switching it by continuously clicking on a single button?

balaji-ramavathu:
@adi23arora I think we need to do that when there are more than 3 sorting options. And I think there's no need to add any more sorting ways in the future.These three options are enough. What do you say? @csoni111 ?

csoni111:
Well there can be more than 3 options. Currently all options are sorting in descending order, we may want to include ascending orders too in future but for now I guess this is fine, we can look for the UI to display all sorting options later on.

balaji-ramavathu:
@csoni111 Yeah I agree. Also, I'm not able to find vector image for the same icon. But I found a similar one which looks like this. Can you tell me if this is okay ?
sort-descending
(Click above to see the image. Won't be visible here since the icon is white :-P )

csoni111:
Yes that is fine. You can try this too, its almost similar to the original one. 
https://visualpharm.com/assets/950/Descending%20Sorting-595b40b85ba036ed117dc061.svg

balaji-ramavathu:
@csoni111 thanks :) I have used your link and changed it to vector drawable.

PR #448, opened 2019-03-20
Fix: Inform user if there's no internet connection while trying to ac…

### LINKED ISSUE #447: Inform user if there's no internet connection while trying to access media files
**Description**
If there is no internet connection while trying to see image files or play any video / audio files, the progressbar is shown continuously instead of displaying an error toast and going back.

**Expected behavior**
A toast should be displayed that there's no internet connection and need to go back to previous page.

balaji-ramavathu:
…cess media files

Fixes #447 

**Summary**
1. For all file types, a toast is displayed that there's no internet connection and goes back to files list page.
2. For video files, if internet is disconnected while the video is playing, EndReached Event is fired and hence the check for internet connection is implemented there.

Please Add Screenshots If any UI changes.

For audio (same for all other types) :

amahi-audio

If internet goes off while playing video :

amahi-video

### DIFF
--- a/src/main/java/org/amahi/anywhere/activity/NavigationActivity.java
+++ b/src/main/java/org/amahi/anywhere/activity/NavigationActivity.java
@@ -173,10 +173,11 @@ public class NavigationActivity extends AppCompatActivity implements DrawerLayou
         if (!CheckTV.isATV(this)) setUpNavigationFragment();
 
         if (isNavigationDrawerAvailable() && isNavigationDrawerRequired(state)) {
-            showNavigationDrawer();
+            this.navigationTitle = getString(R.string.title_shares);
+            setUpTitle();
+            setUpShares();
         }
 
-        setUpNavigationTitle(state);
     }
 
     private void setUpNavigationDrawer() {
@@ -222,14 +223,6 @@ public class NavigationActivity extends AppCompatActivity implements DrawerLayou
     @Override
     public void onDrawerStateChanged(int state) {
         navigationDrawerToggle.onDrawerStateChanged(state);
-    }
-
-    private void setUpNavigationTitle(Bundle state) {
-        this.navigationTitle = getNavigationTitle(state);
-
-        if (isNavigationDrawerAvailable() && !isNavigationDrawerOpen()) {
-            setUpTitle();
-        }
     }
 
     private String getNavigationTitle(Bundle state) {


src/main/java/org/amahi/anywhere/activity/ServerFileAudioActivity.java
```diff
@@ -161,6 +163,12 @@ private void prepareFiles() {
             serverFiles.add(serverFile);
             getIntent().putExtra(Intents.Extras.SERVER_FILES, new ArrayList<Parcelable>(serverFiles));
         }
+
+        NetworkUtils networkUtils = new NetworkUtils(this);
+        if (!networkUtils.isNetworkAvailable()) {
+            Toast.makeText(this, R.string.message_connect_and_try_again, Toast.LENGTH_SHORT).show();
+            finish();
+        }
```
shivammaindola: I think it will be better if you check the connection before opening the activity.

shivammaindola (changes requested):

src/main/java/org/amahi/anywhere/activity/ServerFileAudioActivity.java
balaji-ramavathu: Yeah I agree. Will change it.

codedentwickler:
Great idea. It is good to have such indicator messages on the app. Maybe we can do better with the messages and the checks though.

There is a possibility the user is running a local server which means he won't need to be connected to the internet to access his files. (Basically accessing files on a local server via a Wifi network.)

This messages might come up at the wrong time in this case. However there is still a good chance, the Wifi network gets disconnected and the files on the local server cant be access. 

We will need to have appropriate checks and messages for these situations.

balaji-ramavathu:
hey @codedentwickler how do I check if the user is running a local server?

codedentwickler:
Yes, An easy way is to do a check like in the snippet below. You can understand how auto, remote and local connections is used by observing how this shared preference key `preference_key_server_connection` is modified.

`    private boolean isConnectionLocal() {
        SharedPreferences preferences = PreferenceManager.getDefaultSharedPreferences(getActivity());
        String preferenceConnection = preferences.getString(getString(R.string.preference_key_server_connection), null);
        return preferenceConnection.equals(getString(R.string.preference_key_server_connection_local));
    }`

codedentwickler:
@csoni111 @cpg what do you think about this feature.

balaji-ramavathu:
@codedentwickler thanks! I will add these checks :-)

balaji-ramavathu:
@codedentwickler I've added checks and messages for the local server case.
Can you have a look once? : )

Akash-Ramjyothi:
This is a really good feature and works fine on testing on both local as well as on remote connections, I recommend @cpg to Merge this PR.

chirag-jn:
Something wrong happened. Some issues while merging instead closed the PR.

PR #449, opened 2019-03-21
Show shares on the home screen

### LINKED ISSUE #423: blank screen when back button presses on navigation activity
when back button is pressed twice (1st to remove nav. drawer,2nd to exit app) after running the app from android studio, a blank screen appears.It only appears if the app is launched from android studio

1. Run the app from android studio,select device
2. on the default activity(navigation activity) press back twice

**Expected behavior**
App should close

ggu

 - Xiaomi mi A2
 - OS: Pie

shivammaindola:
Fix #423  #411 
Here we are showing the list of shares on the home screen when the app is launched as suggested by @octacode .

video2gif_20190321_172723


codedentwickler:
Great feature. Can this be merged? @octacode @csoni111 @cpg .

Akash-Ramjyothi:
I have tested this PR and works fine. It does not make much of a difference, but a good feature.

PR #452, opened 2019-03-22
Add register button in login activity

shivammaindola:
For the new user, there is no direct create account option in the login page. It would be better if there is an option like "Create an account" on the login page as shown in gif.

Register

### DIFF
--- a/src/main/java/org/amahi/anywhere/activity/AuthenticationActivity.java
+++ b/src/main/java/org/amahi/anywhere/activity/AuthenticationActivity.java
@@ -23,7 +23,9 @@ import android.accounts.Account;
 import android.accounts.AccountAuthenticatorActivity;
 import android.accounts.AccountManager;
 import android.app.Activity;
+import android.content.Intent;
 import android.graphics.PorterDuff;
+import android.net.Uri;
 import android.os.Bundle;
 import android.support.design.widget.TextInputLayout;
 import android.support.v4.content.ContextCompat;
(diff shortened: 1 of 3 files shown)


shivammaindola:
@cpg @csoni111 can you please review this PR?

adi23arora:
For using Amahi, creating an account is not the only thing. You have to setup your own Amahi Server (HDA). I guess that's the reason why this "register" option was not present in the first place.
Moreover, even if this is implemented, it's always better to use CustomTabs for opening URLs inside the app.

shivammaindola:
I know but if one knows only about the app nothing else then the first thing he'll do is download the app from play store. In that case,  it would be better if there is an option of register on the login page.
Thanks for the suggestion. I will implement CustomTabs.

codedentwickler:
I think having a clue how to setup Amahi from the app is a good idea. But having a Button is misleading as it implies an account can be setup on the app. 
I also would not advice trying to do any account creation on the app as this might lure the user to thinking they don't need to install a server or something.

What I think is the best approach would be to direct a new user to Amahi Overview page where they can learn about Amahi and hopefully setup a server on their Home PC.

Something like a link that says, 

<New to Amahi, Please checkout how you can setup your Amahi server> <link to Amahi website embedded.>

@cpg has been busy of late, He'll give the go ahead on this feature soon, this is what I think is best so we don't mislead potential users.

shivammaindola:
Yeah, I think you are correct. Register button on the login page can be misleading. I will send PR soon.

shivammaindola:
@codedentwickler @cpg @adi23arora Please tell me how is this?

video2gif_20190325_141344

chirag-jn:
@cpg this seems to be a good addition. Can we merge this?

cpg:
It's not constructive to have users of the app that may be noobs register and never install an HDA, because that's not light works.

cpg:
Sorry about the closing. It's a shame we could not use the great work you guys did @shivammaindola and the rest.

PR #460, opened 2019-04-04
Add Fingerprint lock feature

### LINKED ISSUE #453: Add an option to ask for password or pattern everytime application start
To make our application more secure, we can add the privacy feature in settings where a user can set his own password or a pattern to unlock the app during startup. User can disable or enable this feature from the settings at any time.

It will look like this.

lock

shivammaindola:
Fix: #453 
Added fingerprint lock feature with a switch in settings to disable or enable the feature-

Screenshot_2019-04-05-00-52-30-655_org amahi anywhere

**Advantages of Using Fingerprint Authentication**
1. Doesn’t matter how sick you are or unable to recollect things, your fingerprint still stays faultless as your identity and can never be misplaced.
2. Fast, Convenient and Reliable to use.
3. Unique fingerprints assure that it’s unlocked just by you.

You can check the full gif [here](https://drive.google.com/file/d/1Suhax2qBfW-KWAfFTzOS_aCURmrtNejn/view)


src/main/res/values/strings.xml
```diff
@@ -181,5 +181,18 @@
     <string name="pause_the_audio" translatable="false">"Pause the audio"</string>
     <string name="upload" translatable="false">"Upload"</string>
     <string name="sign_in" translatable="false">"Sign in"</string>
+    <string name="upload_channel" translatable="false">File Upload</string>
```
chirag-jn, line 184: There seems to be an error in the upload_channel string value according to the travis build. Though, I don't see any error here in the file.

shivammaindola: Yeah, There is no error. @csoni111 can you help me?

csoni111: @shivammaindola The string `upload_channel` is already present in the current beta branch https://github.com/amahi/android/blob/beta/src/main/res/values/strings.xml#L185 , hence it getting duplicated here.
Merge latest beta branch in your local branch and you will get the error.
Also why have you added these strings here? It is not related to the functionality you are implementing.

src/main/java/org/amahi/anywhere/util/FingerprintCheck.java
```diff
@@ -0,0 +1,79 @@
...
+import android.support.v4.app.ActivityCompat;
+import android.support.v4.content.ContextCompat;
+import android.widget.TextView;
+
+import org.amahi.anywhere.R;
+import org.amahi.anywhere.activity.NavigationActivity;
+
+
+/**
+ * Created by whit3hawks on 11/16/16.
```
csoni111, line 21: Please don't include any personal names here, instead give the documentation here for what is the purpose of this class.

src/main/java/org/amahi/anywhere/util/FingerprintCheck.java
```diff
@@ -0,0 +1,79 @@
...
+ */
+@RequiresApi(api = Build.VERSION_CODES.M)
+public class FingerprintCheck extends FingerprintManager.AuthenticationCallback {
```
csoni111, line 24: `FingerprintManager.AuthenticationCallback` this class is deprecated. Please update to the currently used implementation.

src/main/java/org/amahi/anywhere/util/FingerprintCheck.java
```diff
@@ -0,0 +1,79 @@
...
+
+    @Override
+    public void onAuthenticationSucceeded(FingerprintManager.AuthenticationResult result) {
+        Intent intent = new Intent(context, NavigationActivity.class);
+        context.startActivity(intent);
+        this.update("Fingerprint Authentication succeeded.", true);
+    }
+
+    public void update(String e, Boolean success) {
+        TextView textView = (TextView) ((Activity) context).findViewById(R.id.errorText);
```
csoni111, line 71: `R.id.errorText` does not exist anywhere in the layouts.

src/main/java/org/amahi/anywhere/activity/FingerprintActivity.java
```diff
@@ -0,0 +1,170 @@
...
+    private Cipher cipher;
+    private TextView textView;
+    private Context context;
+
+
+    @RequiresApi(api = Build.VERSION_CODES.M)
+    @Override
+    protected void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        setContentView(R.layout.activity_fingerprint);
```
csoni111, line 55: This layout does not exist.

csoni111 (changes requested):
Seems like you have missed adding a layout file.

csoni111:
I think it would be good idea to add a fallback pass code on the fingerprint login screen. At times your finger may be dirty or just that your fingerprint might not work.

Webp net-resizeimage
Reference: [Splitwise](https://play.google.com/store/apps/details?id=com.Splitwise.SplitwiseMobile)

It will be only enabled if the user has enabled fingerprint lock in settings.
Also add a logout button on the fingerprint login screen.

src/main/res/values/strings.xml
shivammaindola: Yes, I forgot to rebase the beta branch. Sorry, I'll fix it.

shivammaindola:
Thank you for the suggestion, I'll add it too.

shivammaindola:
Since 15 days I was suffering from fever that's why I couldn't worked on it. Now I'm alright and would like to continue this. 
As I messed up some things in this branch, so I'm closing this and opening a new PR.
You can check this PR #473

PR #463, opened 2019-04-09
Fix: Query in the Search bar is not preserved on screen rotation

### LINKED ISSUE #462: Query in the Search bar is not preserved on screen rotation
**Describe the issue**
In ServerFilesActivity, on screen rotation to landscape or from landscape to portrait mode, the activity is recreated, hence any query in the search bar gets lost.

1. enter any search query
2. Rotate to landscape

**Expected behavior**
Search query should not be lost.

**Screenshots**
orientation-gone

I am going to fix this issue.

balaji-ramavathu:
Fixes #462 

The ServerFilesActivity is recreated on screen rotation, which is being done automatically by the system. But here we don't need to recreate the activity, so on adding configChanges property to the activity, we restrict the activity to not recreate due to the phone orientation change.

**Screenshots**
orientation-came

**Summary**
Added configChanges property to the ServerFilesActivity so that it won't recreate on screen rotation.

### DIFF
--- a/src/main/java/org/amahi/anywhere/fragment/ServerFilesFragment.java
+++ b/src/main/java/org/amahi/anywhere/fragment/ServerFilesFragment.java
@@ -132,6 +132,7 @@ public class ServerFilesFragment extends Fragment implements
     private ProgressDialog deleteProgressDialog;
     private int deleteFilePosition;
     private int lastSelectedFilePosition = -1;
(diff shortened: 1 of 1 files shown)


codedentwickler:
Nice catch @balaji-ramavathu. @cpg I think we can merge this. Minor fix here.

PR #469, opened 2019-04-13
Fix: Alert dialogs doesn't persist on screen rotation

### LINKED ISSUE #468: Alert dialogs doesn't persist on screen rotation
**Describe the issue**
Alert dialogs like Delete file confirmation dialog, Duplicate file upload dialog disappear on screen rotation.

1. Choose Delete option of any server file or any recent file / try to upload any file which already exist.
2. Rotate device when the dialog pops up
3. See that dialog disappear

**Expected behavior**
Dialog should not disappear.

I am going to work on this.

balaji-ramavathu:
Fixes #468 

**Screenshots**

For duplicate file upload dialog:

alert-dialog1

For delete file dialog in both ServerFilesFragment, RecentFilesActivity:

alert-dialog2

**Summary**
The alert dialogs are changed to dialog fragments
Implementing alert dialogs as dialog fragments is an easier way to prevent them from disapearing when the screen orientation is changed.

### DIFF
--- a/src/main/java/org/amahi/anywhere/activity/RecentFilesActivity.java
+++ b/src/main/java/org/amahi/anywhere/activity/RecentFilesActivity.java
@@ -14,7 +14,6 @@ import android.support.annotation.NonNull;
 import android.support.annotation.RequiresApi;
 import android.support.design.widget.Snackbar;
 import android.support.v4.widget.SwipeRefreshLayout;
-import android.support.v7.app.AlertDialog;
 import android.support.v7.app.AppCompatActivity;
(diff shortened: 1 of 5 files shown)


codedentwickler (changes requested):
I think it is an overkill to have a new fragment for each alert dialog. I think a better approach is a single fragment that extends `DialogFragment` and it is made configurable. Basically you can have setters or a way to configure it. 

With your current implementation you have too many duplicate code.

balaji-ramavathu:
I messed up some things in this branch, so I'm closing this and will open a new PR

PR #472, opened 2019-04-20
Fix: Alert dialogs doesn't persist on screen rotation

### LINKED ISSUE #468: Alert dialogs doesn't persist on screen rotation

balaji-ramavathu:
Fixes #468

**Screenshots**

For duplicate file upload dialog:

duplicate-dialog

For delete file dialog (both ServerFilesFragment, RecentFilesActivity):

delete-dialog

**Summary**
Implementing alert dialogs as dialog fragments is an easier way to prevent them from disappearing when the screen orientation is changed. So the alert dialogs are changed to AlertDialogs.

A single AlertDialogFragment is created to handle all the alert dialogs. Right now delete file, duplicate file dialogs are implemented.

For delete file dialog:
The index of selected item from the list of files is preserved using onSaveInstanceState, so as to preserve the selected item on screen rotation (in both ServerFilesFragment and RecentFilesActivity).

### DIFF
--- a/src/main/java/org/amahi/anywhere/activity/RecentFilesActivity.java
+++ b/src/main/java/org/amahi/anywhere/activity/RecentFilesActivity.java
@@ -14,7 +14,6 @@ import android.support.annotation.NonNull;
 import android.support.annotation.RequiresApi;
 import android.support.design.widget.Snackbar;
 import android.support.v4.widget.SwipeRefreshLayout;
-import android.support.v7.app.AlertDialog;
 import android.support.v7.app.AppCompatActivity;
 import android.support.v7.widget.DividerItemDecoration;
 import android.support.v7.widget.LinearLayoutManager;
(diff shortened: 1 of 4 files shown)


balaji-ramavathu:
@codedentwickler as you suggested in the previous PR, I have implemented a single fragment that extends DialogFragment and it handles all the alert dialogs. Based on the 'dialogType' that is passed to the fragment, the corresponding dialog is displayed. Can you please review and suggest any other  changes required ?

cpg:
Please indicate what bugs does this close and close them.

balaji-ramavathu:
This fixes #468 only, and it is closed.

PR #492, opened 2019-07-29
Resolved merge conflicts in #491

balaji-ramavathu:


csoni111:
@balaji-ramavathu can you check why the Travis is failing?

balaji-ramavathu:
@csoni111 fixed it

csoni111:
Getting this error on installing the app over the previous version (in which I was logged in):
```
2019-07-31 01:10:46.155 2703-2703/org.amahi.anywhere E/AndroidRuntime: FATAL EXCEPTION: main
    Process: org.amahi.anywhere, PID: 2703
    java.lang.NullPointerException: Attempt to invoke virtual method 'boolean java.lang.String.equals(java.lang.Object)' on a null object reference
        at org.amahi.anywhere.fragment.NavigationFragment.run(NavigationFragment.java:198)
```
My guess is the account that was previously stored has no `is_local` key in it, so its returning null. Hence the error.

chirag-jn:
Since #491 got closed, I guess we can close this PR too.

cpg:
N/A any more.

PR #494, opened 2019-08-16
feat: music player shuffle and repeat features

### LINKED ISSUE #323: Add shuffle to player
A useful feature for the (already excellent) audio player would be to add random shuffle to play files in a given folder in random order over time.

This would allow users to have folders for specific music that they like (e.g. workout music) but not having it in the same order.

I am not sure what is the best UI. Maybe pressing long on the play button offers two options? 1) option `Play in Order` to play in the current sort order for the folder or 2) `Shuffle` to play in an order that is random forever (maybe 3) `Cancel`?)

### LINKED ISSUE #493: Music player Shuffle and Repeat features
shuffle and repeat options in the music player can provide better audio streaming experience to the user. So the music player should have:
1. Shuffle toggle button that can be turned on/off anytime and when it is turned on, audio files will be  shuffled to some random order.
2. Repeat option should toggle between repeat all and repeat only one. For repeat all option, the audio list will be repeated and for repeat only one option, only the current playing song is looped.

balaji-ramavathu:
Fixes #493  
Fixes #323 

**Changes made:**
1. The shuffle option toggles between shuffle on and off, everytime the shuffle mode is changed, it is stored in shared preferences and if the shuffle mode is on, the audio files will be shuffled before going to the media source.
2. The repeat option toggles between repeat all and repeat one As repeat all is already implemented, repeat only current file is made by using LoopingMediaSource.

**Screenshots**

shuffle_rep_res


chirag-jn:
Fixes #323

chirag-jn:
Proposing PR #576 to fix merge issues.

Issue #496, opened 2019-09-09
Showing connection error instead of wrong credentials on sign in page.


mr-volatile:
![Screenshot_20190910-002617](https://user-images.githubusercontent.com/26926669/64558714-29059c00-d362-11e9-802b-d4b7ed30b262.png)

mr-volatile:
And provide correct login credentials.

mr-volatile:
Why there is no Signup/Register option??

cpg:
Good point -- it should be pointed out in the app that the Amahi app needs a home server to be used. We put that in the description (IIRC; well, also in iOS, since I remember putting it), but it probably would be better in the app.

Also, the error should be corrected.

vaandhare:
Thank you @cpg.
Can you please provide me correct credentials to access the app?

cpg:
This is not the place to request them. Follow the details in the README please

mr-volatile:
Hello @cpg 
This was my first PR, please can you merge this PR?

balaji-ramavathu:
Hey @vaandhare you need to set API properties before building the app. You need to follow steps in README to do that. Seems like you haven't done that. This is the reason you are getting inappropriate error message. 

The app contacts with API via the internet to check the credentials. So it first checks if there is internet connection and only if the connection is available, it checks if the credentials are correct or not. For that, API properties need to be set.

shrootis:
Same problem is with me..though net is on there is connection problem...So I am mailing to support team for url and details..

shrootis:
This is something to keep private and you obtain it by sending a message to support at Amahi dot org.
I mailed but didnt get reply....is it support@amahi.org only na

somenath1435:
@octacode @cpg @shrootis @balaji-ramavathu How to get secret key? Please elaborate

spielers:
@somenath1435 message to support@amahi.org about credentials 
they will give 
url.amahi = URL
url.proxy = URL
client.id = ID
client.secret = SECRET
chromecast.app.id = APP_ID
 hoe this will help you

somenath1435:
@spielers I've already sent an email to support@amahi.org but haven't got any reply

spielers:
@somenath1435 ask in IRC Channel

somenath1435:
@cpg @octacode @balaji-ramavathu please suggest what to do

PR #497, opened 2019-09-09
Signin screen showing wrong error

mr-volatile:
Fixed: issue #496 

Sign in screen shows "Connection error. Check your settings and try again." instead of "The username or password are not correct. Please try again." after putting wrong credentials.


Akash-Ramjyothi:
This PR is invalid, I would recommend @cpg to reject this and close it.

PR #498, opened 2019-10-03
Update activity_recent_files.xml

sanjiv1705:


chirag-jn:
Since the app has both light and dark theme, hardcoding the colors in xml might not be the best option. @cpg We can close this PR.

PR #499, opened 2019-10-23
updated orientation

Ruchi110:
minute change


octacode:
We don't want to lock our orientation to vertical all the time. For example: the medial player runs in landscape orientation. Closing this for now.
Thanks!

PR #500, opened 2019-11-22
This now allows an option for Sign up

sirAdarsh:
I've added an option for direct Sign-up for Amahi using ACTION_VIEW intent.


chirag-jn:
Duplicate of #452

cpg:
Sorry we cannot use this good work. We do not want accidental users to be signing up and never install an Amahi system. That is paramount to Amahi. Thanks for your contribution.

PR #501, opened 2020-01-03
Crash Solved Due to Storage and camera Permissions and Added Dialogue box 

samarth30:
# crash due to permissions
Crash solved happening due to Storage and camera permission 
In the below video the app is crashing when i denied the permission one time and then clicked the camera or storage there was no dialogue to seek for permission so the app crashed

<a href="https://imgflip.com/gif/3l2o6o"></a>
# solution
Now i solved the crash by a making a dialogue box and also added your snackbar properties on it that were set earlier

<a href="https://imgflip.com/gif/3l2oe9"></a>

# also added
but i also added that if the user denied to give permission but at the end he wants to upload photo 
so he will see a dilogue box that will take hinm to the settings to allow the app for permissions
this is only case if he denies and you can aslo see this in the video

# video

lastly

# steps to get crash
if you want to get look in to the crash then =
1 click on share on navigations bar
2 click on any of the list item
3 click on the floating button and then on camera
4 deny the permission
5 then again click on the floating button in camera
6 you will see a white screen and this is the crash

## phone details 
my phone is oppo realme with updated android version 9.0 (nugget) 

hope you will merge this pull request @cpg

PR #504, opened 2020-01-06
FEAT: Music Player UI Enhanced

samarth30:
# Enhancement Of UI
here i enhanced ui i turned the toolbar and the exo player background to transparent so that it will increase the user experience
Screenshot_2020-01-06-13-03-36-21_8f7f52555d974a92fa29728873f61978

Screenshot_2020-01-06-13-04-47-92_8f7f52555d974a92fa29728873f61978

# Animation
here i also added animation when the player open so that it will too increase the user experience here i have used animatoo library the changes can be seen in the vdieo 

asdf

phone details
my phone is oppo realme with updated android version 9.0 (nugget)

hope you will merge this pull request @cpg


chirag-jn:
Suggestion Needed: Do we need this change? @cpg

src/main/AndroidManifest.xml
```diff
@@ -31,7 +31,8 @@
     <uses-permission android:name="android.permission.MANAGE_ACCOUNTS" />
     <uses-permission android:name="android.permission.USE_CREDENTIALS" />
     <uses-permission android:name="android.permission.WAKE_LOCK" />
-    <uses-permission android:name="android.permission.WRITE_SETTINGS" />
+    <uses-permission android:name="android.permission.WRITE_SETTINGS"
```
chirag-jn: Ignoring linting warnings in this way might not be the best thing to do.
Ref: https://riptutorial.com/android/example/1135/using-tools-ignore-in-xml-files

chirag-jn:
Too many things happening in this PR. You are handling permissions issue in this PR too which was happening in another PR at the same time. Ref: #501

chirag-jn:
@samarth30 this PR contains your code from all other PRs too which is very difficult to manage since a change in one PR will force to make changes in all other PRs of yours. Please check the same as this is leading to unnecessary conflicts.

chirag-jn:
Swipe Left animation goes against the app's theme of blink animation.

PR #508, opened 2020-02-17
Some minor changes that could be worked for more efficiency of program.

shrootis:
this is my attempt trying to contribute as a beginner,would be highly grateful if mentored on this project and initiative.As seven small guidelines matter and I can play best part of my role.


chirag-jn:
It's a great start @shrootis. However, these fixes have little to no impact on efficiency. Instead, they come under code refactoring.

PR #526, opened 2020-03-20
should be using handler in splash screen

prince-09:
The splash screen should be using a handler for a better experience.


prince-09:
@cpg  please review.

chirag-jn:
There is no need to add a handler in the splash activity. It changes as soon as the next activity starts. So the timing is dynamic.

chirag-jn (commented):
The delay isn't necessary.

cpg:
Closing based on recommendation, as the next activity takes care of it.

PR #533, opened 2020-06-12
Fix Translations in strings and keys

### LINKED ISSUE #430: Removal of hard coded strings

chirag-jn:
Fixes #430 
Extends #432


cpg:
Anything that is not translatable should really not be in any strings file, no?

e.g. `DownloadFinished`, or `INTENT_FILE` or `auto_upload_server`

correct?

chirag-jn:
Created a Constants class as recommended by @codedentwickler in #432 
Creating a `static final` variable doesn't cause any type of memory leak when compared to `static` variables which are potential points of memory leaks.

PR #516, opened 2020-03-06
FEAT : Animation in RecyclerView Added  

samarth30:
#428 solved

Added Animation in Recycler view in Recent files and in server files adapter. this will give the more better user experience and enhance the UI

# VIDEO
<a href="https://imgflip.com/gif/3rje3a"></a>

could you please review this @cpg @octacode

PR #438, opened 2019-03-12
Added animation to RecyclerView

shivammaindola:
Added fade-in type of animation to recyclerView. #428

PR #539, opened 2020-06-19
Animation in RecyclerView

### LINKED ISSUE #428: Implementing Animation in RecyclerView
video2gif_20190306_181602

@cpg @octacode Can I add animation in above list while scrolling? I think, it will look preety cool.

chirag-jn:
Extends #438 #516 
Fixes #428


chirag-jn:
The current changes implement animations at an item level and not view level, which might not lead to a good experience in case of long lists.

PR #544, opened 2020-06-28
Added support for VPN Proxy

### LINKED ISSUE #547: VPN Proxy Tunnel Support
**Feature Request:**

- Add support for VPN proxy so that user can access the HDA from the phone over a VPN proxy tunnel

- This would establish a tunnel to VPN Server to replace regular ISP routing and to encrypt and secure all network traffic.

- To enable multi-hops on connections through the server.

- To prevent privacy leaks while accessing server files.

- To prevent Geo-Blocks.

- To reduce Bandwidth throttling.

### LINKED ISSUE #559: Unable to connect to Amahi home server when not on my home network
I am unable to connect to my share and file in my server when not on my home network,  I remember this work a few years ago , I have test this on the version of the app on google store and the gsoc2020 version too, I am running Amahi 11 on fedora 25 on my live server not  a vm machine, 

**Expected behavior**
No connection 

**Screenshots**

### LINKED ISSUE #596: VPN Issue with the app
**Describe the issue**
When I am not at home  I can use open vpn to connect to my server this as been working great for the last few years and it still work, so from outside my network I can see all my shares and the Amahi front end or dashboard, I can run apps on my server this is great but I am unable to use the android app, I know amahi anywhere is have problems,  but if I have connected with a vpn it should use this frist , if I was in a coffee stop was using a vpn for security over the network I would have thought the app was using the vpn too

1. Setup openvpn on your server

**Expected behavior**
Should just work

**Screenshots**

 - Device: all devices 

**Additional context**
I am able to test this

Akash-Ramjyothi:
The VPN proxy tunnel is established between the client app and the server when the "VpnProxyTunnel.java" is initialized and its VPN service has been added on the manifests of the Server Files while accessing them over the HDA which is always set On when the user authenticates and accesses the HDA.

### DIFF
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -97,6 +97,12 @@
                 android:name="android.app.searchable"
                 android:resource="@xml/searchable" />
         </activity>
+        <service android:name=".activity.ServerFilesActivity"
+            android:permission="android.permission.BIND_VPN_SERVICE">
+            <intent-filter>
+                <action android:name="android.net.VpnService"/>
+            </intent-filter>
+        </service>
         <activity
             android:name=".activity.ServerFileAudioActivity"
             android:launchMode="singleTop"
--- a/src/main/java/org/amahi/anywhere/util/VpnProxyTunnel.java
+++ b/src/main/java/org/amahi/anywhere/util/VpnProxyTunnel.java
@@ -1,0 +1,90 @@
+package org.amahi.anywhere.util;
+
+import android.app.NotificationManager;
+import android.app.PendingIntent;
+import android.content.Intent;
+import android.net.VpnService;
+import android.os.Handler;
+import android.os.Message;
+import android.os.ParcelFileDescriptor;
+import android.util.Log;
+import android.util.Pair;
+import android.widget.Toast;
+
+import org.amahi.anywhere.activity.AuthenticationActivity;
+
+import java.io.IOException;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
+public class VpnProxyTunnel extends VpnService implements Handler.Callback {
+    private static final String TAG = VpnProxyTunnel.class.getSimpleName();
+    public static final String ACTION_CONNECT = "com.example.android.toyvpn.START";
+    public static final String ACTION_DISCONNECT = "com.example.android.toyvpn.STOP";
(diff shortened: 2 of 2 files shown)


cpg:
We need rationale of why this is needed, and how it's done, the reliability of the service and a diagram of how this is happening in what you propose.

Akash-Ramjyothi:
Fixes #547 

- **How it's Done:**
The VPN Tunnel works on the protocol "OpenVPN" which is a newer version of SSTP - Secure Socket Tunneling Protocol. This is implemented through the ToyVpn Client Service.  It is done when a connection is established between the client app and the server that surrounds the network data like a tunnel, encrypting the data packets that the app sends. Tunneling is the process of encrypting data and keeping it separate from other traffic on the internet. It enables a VPN to protect the confidentiality and integrity of data as they travel over the public network. The VPN proxy tunneling protocol uses AES 256-bit encryption to protect data packets. The protocol is configurable on Windows, Mac, Android, and iOS, although third-party client service (ToyVpn) is required to set up the protocol.

- **Reliability of the Service:**
Among all the tunneling protocols, OpenVPN is the most highly recommended. OpenVPN works with systems like FreeBSD, NetBSD, OpenBSD, and Solaris. Using VPN alone may not be enough to protect privacy and that is the reason VPN service providers encrypt the data and receive when connected. When this is done, it scrambles so that snoops can't intercept and decipher it. Its open-source code, stability, strong encryption, and ability to bypass blocking software make it the finest tunneling protocol currently available.

- **Pictorial Representation:**
Regular Connection with no VPN:
novpn

Traditional VPN:
tradvpn

VPN Tunnel Implemented:
what-is-vpn-diagram

- **How this is Happening:**
It works by routing your device’s internet connection through your chosen VPN rather than your internet service provider (ISP) so that when the data is transmitted to the internet, it comes from the VPN rather than the device. The VPN acts as an intermediary of sorts to connect to the internet, thereby hiding the IP address – the string of numbers the ISP assigns to the device. And if the data is somehow intercepted, it will be unreadable until it reaches its final destination.
When it connects to a VPN, it is through a secure tunnel where the data is encoded. This means that the data is transformed into an unreadable code as it travels between the device and the server.

chirag-jn:
How do we exactly decide if we have to change the connection mode to `VPN`? Or is it like it will always connect via `VPN`? In my opinion, that will unnecessarily throttle our connection with the server.

PR #553, opened 2020-07-06
Fix: Inform user if there's no internet connection while trying to access media files

### LINKED ISSUE #447: Inform user if there's no internet connection while trying to access media files

chirag-jn:
Fixes #447
Extends #448 

Summary

For all file types, a toast is displayed that there's no internet connection and goes back to files list page.
For video files, if internet is disconnected while the video is playing, EndReached Event is fired and hence the check for internet connection is implemented there.

image

If internet goes off while playing video :

image


chirag-jn:
Moved to a Snackbar instead of a Toast message
image

PR #560, opened 2020-07-12
fix change between light and dark them settings app crash

Ahmed-Abd-Elsamie:
@chirag-jn  I have figured out the problem and fixed it.
the problem was in the NavigationActivity.class , after changing between light and dark theme we finish the activity then start it again from itself. and this will cause the app to crash because we are trying to start it from null context.

Solution :
i removed startActivityForResult function and replaced it with the following:
1 - once opening the settings page i finish the navigationActivity.
2 - once do any change in the settings like changing between light and dark themes, we start the NavigationActivity again from the settingActivity.

I tried this method and solved the problem

here is the updated

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -110,6 +110,8 @@ static def formatStringField(field) {
 }
 
 dependencies {
+    implementation 'androidx.appcompat:appcompat:1.1.0'
+    implementation 'androidx.constraintlayout:constraintlayout:1.1.3'
     repositories {
         google()
         mavenCentral()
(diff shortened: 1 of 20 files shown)


chirag-jn:
@Ahmed199810 Please make a fork of the latest `beta` branch since there a lot of conflicts in the files, and there have been enough changes in the codebase in the last few days.

chirag-jn:
Also, please add the code relevant to this PR only. I can see that you have changes present related to NAU and friending activity but those aren't relevant to this PR.

chirag-jn:
Also, if I'm not wrong, this change has been already incorporated in the `beta` branch, please confirm the same.

cpg:
Already in the codebase. Thanks much for your contribution.

PR #563, opened 2020-07-16
Add Non-Admin User Support

### LINKED ISSUE #335: Rights of a secondary user logged into hda?
What would be the rights granted to the secondary user while adding a support for secondary user login? Will it be restricted to viewing rights only of the shared content specified by admin. Or will it have support for admin to select which contents would be made available to a specific secondary user and also its access rights to "read only" or "modify" individually on each secondary user?

### LINKED ISSUE #525: User behavior issue with spaces in Login
**Describe the issue**
When we try to login in the app, if we have whitespace on the right of the username (which weren't supposed to be there), login still works.

1. Enter your username and password which should work fine.
2. Before pressing the sign in button, enter any amount of whitespace on the right. 
3. Try signing in.
4. It should signin successfully.

**Expected behavior**
It should show an error message that "The username or password are not correct."

**Smartphone**
 - Device: Vivo 1807
 - OS: Android 8.1

### LINKED ISSUE #611: NAU testing for release candidate release
**Describe the issue**
We want to test NAU so that the app works on Amahi 12 as well as older Amahi 11 and Amahi 10 server

We can probably test an AAFS v10 on an Amahi 11 VM.

chirag-jn:


chirag-jn:
Issues with auth_token, will be fixed soon.

chirag-jn:
Non-Admin User works fine now. User is able to login by just using a PIN which they can set up at `http://hda/tab/users/`. However, enough user testing is required to check the corner cases,

Akash-Ramjyothi:
@chirag-jn I feel it would be better to replace the normal "EditText" on the PIN Access Fragment and replace them with "TextInputLayout - Floating EditText" with counter enabled and locking it to 4 or 6, alternatively to indicate by a change in color if character exceeds.

src/main/res/layout/fragment_pin_access.xml
```diff
@@ -0,0 +1,90 @@
...
+
+    <EditText
+        android:id="@+id/edit_text_pin"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:gravity="center"
+        android:hint="@string/hint_pin"
+        android:imeOptions="actionGo"
+        android:inputType="textPassword"
+        android:textSize="@dimen/text_large" />
```
Akash-Ramjyothi: These are the codes I recommend you to convert to the UI element I specified.

Akash-Ramjyothi (dismissed):

chirag-jn:
That seems to be a good change, though note that the length of the pin can be 3-5

chirag-jn:
Corrected issue of navigation bar not visible for non admin user.

Issue #558, opened 2020-07-09
Android app force close


sanjitchak:
**Describe the issue**
Android force closing every few seconds, after running in the background for an hour.

**To Reproduce or Feature details**
1. Open the app
2. Keep it on the background
3. Keep doing other stuffs on your phone like browsing web.

**Expected behavior**
No force close

**Screenshots**
 
![unnamed (2)](https://user-images.githubusercontent.com/20724062/87062242-6eb51e80-c22a-11ea-88a0-f6b0b81c073f.jpg)

**Smartphone (please complete as much of the following information as possible):
Device: Asus max pro m1
OS: Android 9

**Additional context**
Maybe add a log collector in the app.

chirag-jn:
The issue which causes app close is as follows:

[...]

PR #564, opened 2020-07-18
Application Crash Reports

chirag-jn:
Adding Support for ACRA (https://github.com/ACRA/acra) which will provide crash logs to developers via email whenever the app crashes.
This logging will happen only in the `debug` mode. It can be easily modified to support `release` mode too. Developers will have to enter their preferred email in `acra.properties` file.


chirag-jn:
This will help us debug issues like #558

PR #591, opened 2020-08-01
Leading and trailing white-spaces of username will be trimmed before it goes for authentication

mitrukahitesh:


chirag-jn:
Fixes #525

chirag-jn:
@cpg should we trim the username before it goes for authentication?

chirag-jn:
Or should we let the server handle that?

mitrukahitesh:
With the commit "Space as first character in username will be avoided", user will not be able to enter a space as first character for username. Clicking space will not move the cursor forward if the username field is empty.

chirag-jn:
@mitrukahitesh You have made changes to the master branch. However, all development is happening on the `beta` branch, and I would recommend you to close this PR, make the same changes on the beta branch and create a new PR.

mitrukahitesh:
@chirag-jn Sure. I will do the same. Thank you.

PR #592, opened 2020-08-02
Avoiding white-spaces in username

### LINKED ISSUE #511: No checks on username TextField if it contains spaces.
In Authentication class there's no condition that checks if username contains spaces before authentication.

mitrukahitesh:
1. White-space will not be accepted as first character in username.

**Before:**
before

**After:**
after

2. Leading and trailing white-spaces in username will be trimmed before going for authentication. Thus server need not check for the same.

Hope that it is useful.

### DIFF
--- a/src/main/java/org/amahi/anywhere/activity/AuthenticationActivity.java
+++ b/src/main/java/org/amahi/anywhere/activity/AuthenticationActivity.java
@@ -74,32 +74,6 @@ public class AuthenticationActivity extends AccountAuthenticatorAppCompatActivit
 
         setUpAuthentication();
 
-        avoidLeadingSpaceInUsername();
-    }
-
-    private void avoidLeadingSpaceInUsername() {
-        getUsernameEdit().addTextChangedListener(new TextWatcher() {
-            @Override
-            public void beforeTextChanged(CharSequence charSequence, int i, int i1, int i2) {
-
-            }
-
-            @Override
-            public void onTextChanged(CharSequence charSequence, int i, int i1, int i2) {
-
-            }
-
-            @Override
-            public void afterTextChanged(Editable editable) {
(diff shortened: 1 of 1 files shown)


chirag-jn:
Fixes #525

mitrukahitesh:
Commit "improvement in removing leading white-spaces from username" enables the following

gg

chirag-jn:
In my opinion, that's over complicating a fairly simple task and we don't want our users to be frustrated on the login screen itself.

mitrukahitesh:
In case a user copies username (with a space at the beginning, which is a common case) and pastes it in the app, will it pass the authentication? If no, then I have a reason to believe that it is useful.

chirag-jn:
@mitrukahitesh the original intent of the PR is excellent where you are eliminating spaces at the beginning. I had concerns regarding the latest commit in this PR. How is it precisely helping?

chirag-jn:
In fact, @mitrukahitesh, if you see this PR thread, you will see that I tagged an issue where I raised the exact same concern a long time back which you are trying to solve here.

mitrukahitesh:
@chirag-jn If you observe carefully, you will notice that the initial commit alone is not sufficient to avoid space(s) at the beginning. For example: If I copy a username with a space at the beginning and paste, then the initial commit alone cannot remove the leading space. Similarly there might be other cases which the initial commit cannot handle alone. Whereas the latest commit takes care of the same. My intention with the latest commit was to assure that at any situation there is no leading space.

mitrukahitesh:
However, if you believe it's not required, I will remove it.

chirag-jn:
@mitrukahitesh It will be merged by tomorrow

cpg:
Just **don't touch the UI please**. That can infuriate users. Just pass things through a cleanup function to remove spaces before and after (trim?).

PR #603, opened 2020-08-08
Remove Timber Dependency

chirag-jn:
`Timber` is an alternate logging library proposed in front of the default `Android.Log` library and just provides a shorthand of using `Timber.d(msg)` in place of `Log.d(TAG, msg)`. 
However, it's rarely used in the codebase and we can remove its dependency to reduce the final apk size.


chirag-jn:
This PR works fine after merging beta.

PR #615, opened 2020-08-11
Update Strings in Preference Screen

chirag-jn:
This is an extension to #608 
Provides translation to the strings in preference screen of uploading photos.

### DIFF
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -230,10 +230,10 @@
     <string name="preference_title_upload_settings">Auto Upload Photos</string>
     <string name="preference_title_language">Language</string>
 
-    <string name="preference_title_upload_server">Select server</string>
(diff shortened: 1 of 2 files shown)


src/main/res/values/strings.xml
```diff
@@ -230,10 +230,10 @@
     <string name="preference_title_upload_settings">Auto Upload Photos</string>
     <string name="preference_title_language">Language</string>
 
-    <string name="preference_title_upload_server">Select server</string>
-    <string name="preference_title_upload_share">Select share</string>
-    <string name="preference_title_upload_location">Upload location</string>
-    <string name="preference_title_upload_data">Allow on cellular data</string>
```
cpg: I was the one that asked that the strings be kept with normal capitalization. I Kind Of Do Not Enjoy Seeing Messages With Capitalization (I think for some things it's too much). If other significant apps do it, I would be open to considering it.

PR #649, opened 2020-08-28
Linting

chirag-jn:
Performed basic linting at few places to remove obscure/redundant pieces of code.


src/main/java/org/amahi/anywhere/activity/RecentFilesActivity.java
```diff
@@ -263,12 +263,6 @@ public void onFileOptionSelected(FileOptionClickEvent event) {
         selectedFileOption = event.getFileOption();
         switch (selectedFileOption) {
             case FileOption.DOWNLOAD:
-                if (Android.isPermissionRequired()) {
```
cpg, line 266: this is more than linting! ...

cpg: Oh, it's because it so happens that the case under it is identical. Hmm..

src/main/java/org/amahi/anywhere/tv/fragment/ServerSelectFragment.java
```diff
@@ -134,8 +134,6 @@ private void setDefaultChecked() {
 
     private void setFalseChecked() {
         if (OPTION_CHECKED != null) {
-            setFalse();
```
cpg, line 137: ok, that `.clear()` may do something ... no?

src/main/java/org/amahi/anywhere/activity/RecentFilesActivity.java
chirag-jn: Yes! Identical cases.

src/main/java/org/amahi/anywhere/tv/fragment/ServerSelectFragment.java
chirag-jn: `OPTION_CHECKED` is an ArrayList (All caps shouldn't be used for variable names though).
`.clear()` removes all elements from that ArrayList. 
However, earlier it could have possibly caused an issue because we were calling the `clear` function while incorrectly checking if `OPTION_CHECKED` is null or not.

cpg: The meaning of the code afterwards is potentially not the same. It will be cleared *if* it's not `null` whereas before it was not.

chirag-jn: I have updated the code.

PR #677, opened 2020-10-01
improved docs

Pranshumehta:


chirag-jn:
Hahaha even though we know it's an amazing app, let's just leave it to the developers to find that out themselves :)

PR #680, opened 2020-10-12
Added Google Login Feature, But Configure Google API console with owner'…

neeleshwark17:
Google Login Feature is added but it will work after configuring Google API console with owner's email.


chirag-jn:
Amahi is a home server. It doesn't require Google login and it follows it's own authentication protocol.

Issue #530, opened 2020-03-31
Navigation Drawer Not Available inside Folders


chirag-jn:
**Describe the issue**
Navigation Drawer isn't available on right swipe inside folders. The burger menu icon at the top gets replaced with a back icon.

**Expected behavior**
For easy navigation, Navigation Drawer should be available on every screen. On swiping right from the edge of the screen, we should be greeted with a navigation drawer.

**Screenshots**
![image](https://user-images.githubusercontent.com/31701616/78047613-31710300-7396-11ea-8f45-588d9ab9e9f3.png)

**Smartphone (please complete as much of the following information as possible):**
 - Device: Vivo 1807
 - OS: Android 8.1

chirag-jn:
If it's a good enough suggestion, can you please let me know if I can start working on this?

Prince-kushwaha:
I want to work on this problems

chirag-jn:
@Prince-kushwaha, sure, give it a try once?

Prince-kushwaha:
actually i am beginner it is my first open source project  i fork the project in my pc when i open it in my android studio it,s activity layout not show. project Gradle and sdk not download automatically 
please help

chirag-jn:
@Prince-kushwaha it would be great if you can post the error you are facing

Prince-kushwaha:
i try but i not solve able to solve this issue

Prince-kushwaha:
i try but i not able to solve this issue

Prince-kushwaha:
tell me i am beginner which issue i chose 
in open issue how know on which issue anyone not  working

Prakhar-Agarwal-byte:
Hey can I work on this issue?

chirag-jn:
> Hey can I work on this issue?

Sure.

PR #685, opened 2020-11-04
Replaced switch with if/else statements for Non-Constant Fields in Ca…

### LINKED ISSUE #683: Non-Constant Fields in Case Labels
**Describe the issue**
With the introduction of Gradle 5.0, resource IDs won't be final anymore. Hence, a switch-case statement won't work in onClick function for view elements. We have to switch to if-else conditions of a view binding strategy for mapping resource IDs to their variables. 

More information: http://tools.android.com/tips/non-constant-fields

For the first case, we have to switch from
``` java
@Override
public void onClick(View view) {

    switch (view.getId()) {
        case R.id.next_btn:
            // Do something
            break;
    }
}
```
to something like this:
``` java
@Override
public void onClick(View view) {
    
    int id = view.getId();
    if (id == R.id.next_btn) {
        nextBtnPressed();
    }
}
```

Old view binding libraries like [JakeWharton/butterknife](https://github.com/JakeWharton/butterknife) won't work anymore either.

Prakhar-Agarwal-byte:
…se Labels
@chirag-jn @cpg 
Please check if any other changes are required if not then please accept the pull request.

### DIFF
--- a/src/main/java/org/amahi/anywhere/activity/NativeVideoActivity.java
+++ b/src/main/java/org/amahi/anywhere/activity/NativeVideoActivity.java
@@ -211,14 +211,11 @@ public class NativeVideoActivity extends AppCompatActivity implements
 
     @Override
     public boolean onOptionsItemSelected(MenuItem menuItem) {
-        switch (menuItem.getItemId()) {
-            case android.R.id.home:
-                finish();
-                return true;
-
-            default:
-                return super.onOptionsItemSelected(menuItem);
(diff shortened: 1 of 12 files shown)


chirag-jn (approved):

Prakhar-Agarwal-byte:
@chirag-jn HELP NEEDED : I am not able to get API properties for building Amahi app. I mailed support@amahi.org but did not get any response. Please let me know how to get those so that I can start contributing more on the this issue ( #530 ).

PR #707, opened 2020-12-10
Removed outdated code which were for Android SDK version lower than 21

### LINKED ISSUE #702: Remove unnecessary files and code for Android version lower than 21
**Describe the issue**

Since we have updated `minSdkVersion` to 21 I think it would be good if we remove code and files that are meant for `minSdkVersion` less than 21. This would also decrease the app size and clean the code.

1. Look for the `minSdkVersion` property in the 'build.gradle' file it is set to **21**.
2. Then lookout for files and code that are meant for android Sdk version lower than `minSdkVersion` 21.

**Expected behavior**

The files and code specifically meant for android sdk version lower than 21 should be removed.

**Additional context**

This is just my opinion please feel free to tell me your suggestions.

Prakhar-Agarwal-byte:
@cpg
I have removed 2 files that were specific to the Android SDK version below 21 and also made some required code changes. Fixes #702


src/main/res/layout-v21/view_server_file_item.xml
```diff
@@ -1,143 +0,0 @@
-<?xml version="1.0" encoding="utf-8"?><!--
```
chirag-jn, line 1: minSDKversion is 21. So why is this file removed?

src/main/res/values-v21/themes.xml
```diff
@@ -1,55 +0,0 @@
-<?xml version="1.0" encoding="utf-8"?>
```
chirag-jn, line 1: Similarly, minsdkversion is 21.

src/main/res/values/themes.xml
```diff
@@ -136,18 +139,28 @@
...
+        <item name="android:statusBarColor">@android:color/transparent</item>
         <item name="windowActionBar">false</item>
         <item name="windowNoTitle">true</item>
+        <item name="drawerArrowStyle">@style/DrawerArrowStyle</item>
         <item name="android:textColor">@color/primary_text</item>
     </style>
 
     <style name="Theme.Amahi.NoActionBar.Audio">
+        <item name="android:windowDrawsSystemBarBackgrounds">true</item>
+        <item name="android:statusBarColor">@android:color/black</item>
```
chirag-jn, line 152: Please attach screenshots for before and after changes in styles/themes

chirag-jn (changes requested):

src/main/res/layout-v21/view_server_file_item.xml
prakharagarwal-dev: Since view_server_file_item.xml (v21) and view_server_file_item.xml (base) are the same, I removed the file with version qualifier so that naming is maintained. It does not matter which file is removed since both files are same.

src/main/res/values-v21/themes.xml
prakharagarwal-dev: I merged both themes.xml (v21) and themes.xml (base), then since explicit v21 qualifier (because minsdkversion is 21 ) is not required I removed themes.xml (v21) file.

src/main/res/layout-v21/view_server_file_item.xml
chirag-jn: Great.

chirag-jn:
Please attach screenshots of the following two things
1. App working on Android v21
2. Changes introduced after the changes in styles/themes files

Prakhar-Agarwal-byte:
1. App working perfectly fine in Android v21
image
image

2. No changes in the theme are introduced. It is the same as before since I have only put extra code from themes.xml (v21) that was specific for v21 and above in the themes.xml (base) file.

src/main/res/values/themes.xml
```diff
@@ -84,13 +87,13 @@
         <item name="android:textColorSecondary">@color/secondary_text</item>
     </style>
 
-    <style name="Theme.Amahi.ActionBar" parent="@style/Widget.AppCompat.ActionBar">
+    <style name="Theme.Amahi.ActionBar" parent="@style/ThemeOverlay.AppCompat.ActionBar">
```
chirag-jn: Can you explain what's happening here?

prakharagarwal-dev: I reverted it.

src/main/res/layout/view_server_file_item.xml
```diff
@@ -18,19 +18,19 @@
...
     android:layout_width="match_parent"
     android:layout_height="wrap_content"
-    xmlns:tools="http://schemas.android.com/tools"
-    xmlns:app="http://schemas.android.com/apk/res-auto"
-    android:background="?android:attr/selectableItemBackground"
+    android:layout_margin="4dp"
     android:clickable="true"
     android:focusable="true"
-    style="@style/Widget.MaterialComponents.CardView"
+    android:foreground="?android:attr/selectableItemBackgroundBorderless"
```
chirag-jn, line 29: Where is the `android:background` attribute? We don't need that anymore?

prakharagarwal-dev: No, it is not required from API 21. We can instead set `android:foreground="?android:attr/selectableItemBackgroundBorderless"`

src/main/res/values/themes.xml
prakharagarwal-dev: Actually, it was present in themes v21 so I changed `Widget.AppCompat.ActionBar` to `ThemeOverlay.AppCompat.ActionBar`. Nonetheless, `Widget` will also work fine.

chirag-jn (approved):
Looks good. Nice job @Prakhar-Agarwal-byte

PR #714, opened 2020-12-20
Changed Highlight color to Accent Color while Searching

### LINKED ISSUE #713: Highlighting Name of File with Accent Color while Searching
**Describe the issue**

Currently, when we search for a file then the file name gets highlighted with #be5e00 color which is not only NOT same as accent color but also it does not change when switching between light and dark mode. It is not consistent with the app color theme as it should match the accent color of the app both in light and dark mode.

1. Go to Shares.
2. Search for a file by typing its name in the search view.
3. See the name of the file getting highlighted in #be5e00 color which is different than accent color bot in light and dark mode.

**Current behavior**

image

image

**Expected behavior**

image

image

 - All Android Devices

Prakhar-Agarwal-byte:
@cpg
I have changed the color by which files get highlighted while searching for them to match the app's accent color, both in dark and light mode. Fixes #713


src/main/java/org/amahi/anywhere/adapter/FilesFilterAdapter.java
```diff
@@ -97,6 +97,14 @@ public void replaceWith(ServerShare serverShare, List<ServerFile> files) {
         notifyDataSetChanged();
     }
 
+    public ForegroundColorSpan getFcs() {
+        if (AmahiApplication.getInstance().isLightThemeEnabled()) {
+            return new ForegroundColorSpan(Color.parseColor("#328ce7"));
```
chirag-jn: Rather move the color codes to `colors.xml`

chirag-jn (changes requested):

chirag-jn:
Rest seems fine.

src/main/java/org/amahi/anywhere/adapter/FilesFilterAdapter.java
prakharagarwal-dev: I tried doing so, but that would require context to be present in the adapter. Since no context is present in `FilesFilterAdapter.java` I chose to hardcode color value.

chirag-jn: You can get the context in an adapter. Check that once. 
https://stackoverflow.com/questions/32136973/how-to-get-a-context-in-a-recycler-view-adapter

prakharagarwal-dev: I checked the link but 
1. `FilesFilterAdapter.java` is an abstract class so I cannot pass context in the constructor.
2. It also does not contain any view from which I can get context.

prakharagarwal-dev: I think I should move `getFcs()` into ServerFilesAdapter and ServerFilesMetadataAdapter then it will work fine

chirag-jn: Yeah try doing that.

prakharagarwal-dev: I have done it.

Prakhar-Agarwal-byte:
@chirag-jn  Please approve this now

chirag-jn (approved):
Looks good.

PR #718, opened 2020-12-28
Addition of github actions

### LINKED ISSUE #717: Introduction to github actions
**Describe the issue**
Github actions is a new way of managing of performing CI/CD and is way faster than Travis since it doesn't have to necessarily download the whole SDK on every build, hence the verification time is much lesser.

Currently travis runs in two stages
1. assembleDebug
2. assembleDebugAndroidTest

We can move both of these tests to Github actions which can provide much faster feedback and since it's free for public repositories, there's no limit to the number of concurrent jobs which is the case with free version of Travis.

chirag-jn:
Fixes #717 
I have been able to run `assembleDebug` on github actions but due to some issue of virtual device, I am not able to run `assembleDebugAndroidTest` on github actions.
Nevertheless, in this PR, I am running `assembleDebug` on github actions and `assembleDebugAndroidTest` on Travis which reduces the time it takes to run the tests. 
In future, we can move `assembleDebugAndroidTest` to github actions too and completely remove the dependency on Travis.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,0 +1,10 @@
+name: Build
+on: [pull_request, push]
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout the code
+        uses: actions/checkout@v2
+      - name: Build the app
+        run: ./gradlew build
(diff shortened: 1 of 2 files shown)


chirag-jn:
Github Actions will be triggered in builds only after its present in the repository tree. Hence, it won't be triggered on this PR but will happen in future PRs.

PR #721, opened 2021-01-05
Added Alphabetical Ordering in Languages in Settings

### LINKED ISSUE #720: Alphabetical ordering to languages in settings
**Describe the issue**
Since we are adding more and more languages to the Amahi Android app, it will be beneficial to add a feature to sort the list of languages in settings alphabetically. In this way, it will be much easier to locate a language for the user as the number of languages supported by the Amahi Android app increases.

1. Go to 'Settings'
2. Click on 'Language'
3. Look at the list of languages which is not in any particular order.
4. This list can be sorted in alphabetical order.

**Current behavior**
List of languages not in any particular order.

image

**Expected behavior**
We should see the list of languages in alphabetical order.

image

 - All Android Devices

**Additional context**
We should implement this feature in such a way so that it can be easily scaled with an increasing number of languages.

Prakhar-Agarwal-byte:
I have added alphabetical ordering in the list of languages in settings. I have implemented it in such a way so that the languages get sorted in runtime only once i.e. when the settings activity is loaded. Also, the languages can be present in any order in the array of entries for the list preference of Languages. So we don't have to worry about ordering in the XML file. Fixes #720


chirag-jn:
However, since English is the preferred option for a vast majority, can we keep it as the first option?

Prakhar-Agarwal-byte:
> However, since English is the preferred option for a vast majority, can we keep it as the first option?

Okay I will add it at top

Prakhar-Agarwal-byte:
@chirag-jn I have added the English language as the first option

PR #725, opened 2021-01-07
Improved Instructions in Languages.md file

### LINKED ISSUE #724: Improve Instructions in Languages.md for adding new Languages
**Describe the issue**
In the [Languages.md](https://github.com/amahi/android/blob/beta/LANGUAGES.md) file 2 points can be improved:

1. In point 1 the value of string should not be "hi-rIN", 
```xml
    <string name="language_hi_key" translatable="false">hi-rIN</string>
```
but it should be "hi", only then it would work.
```xml
    <string name="language_hi_key" translatable="false">hi</string>
```
2. In point 2 we can be more specific, to avoid any confusion, and change the point from
```xml
Add `language_hi` key in every strings file.
```
to
```xml
Add `language_hi` key in every strings file, translated into their respective language. For Example, in `strings.xml (hi)` add ```<string name="language_hi">हिंदी </string```
```

**Additional context**
Many people were commiting the same mistake so I thought we can change the instructions a bit to reduce confusion.

Prakhar-Agarwal-byte:
I have changed the instructions as mentioned in the issue. Fixes #724


chirag-jn (approved):
Looks good.

PR #755, opened 2021-03-19
 Wrong Translation in Hindi

msubham193:
There Is a Wrong Translation In String.xml (hi) Folder at 55 line


cpg:
What is the -od file about? Seems french, and you do not refer to it.

Also tests are failing, and it appears that the errors are due to that file.

msubham193:
@cpg  iam Working On this od file.

